PR #11, opened 2018-08-30
Add bazel test presubmit for testing repo

munnerz:


munnerz:
🤞 this should hopefully update automatically

PR #70, opened 2018-09-07
Remove tarmak config

munnerz:


munnerz:
Pending checking with @simonswine

PR #87, opened 2018-09-18
Adds nodejs dependency and reference docs run for tarmak

JoshVanL:
/cc @simonswine 
/cc @munnerz

### DIFF
--- a/config/jobs/tarmak/tarmak-presubmits.yaml
+++ b/config/jobs/tarmak/tarmak-presubmits.yaml
@@ -548,12 +548,7 @@ presubmits:
         - make
         - -C
         - docs
-        - clean
-        - reference_docs_generate
-        - spelling
-        - linkcheck
-        - html
-        - copy_artifacts
+        - ci
         resources:
           requests:
             cpu: 750m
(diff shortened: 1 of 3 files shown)


config/jobs/tarmak/tarmak-presubmits.yaml
```diff
@@ -549,6 +549,7 @@ presubmits:
         - -C
         - docs
         - clean
+        - reference_docs_generate
         - spelling
```
munnerz: This will break existing PRs until your own is merged, right?

May be better to add this as a separate job with always_run set to false for the time being, so that other PRs don't fail 😄

simonswine: We should have a CI target in the Makefile. The tasks should be added to the tarmak master first and the update this here to run target `make -C docs ci`. After that we can easily update it without testing

simonswine:
Ok we need to update the image used to 
`eu.gcr.io/jetstack-build-infra-images/tarmak-sphinx-docs:v20180919-4a6536b`

Can you do that @JoshVanL

JoshVanL:
done @simonswine

simonswine:
Cheers, it should be ready in about 1min

PR #93, opened 2018-10-22
cert-manager: add kubernetes 1.12 non-blocking presubmit

munnerz:


munnerz:
GitHub webhooks currently not working due to github outage so...
/hold

PR #112, opened 2018-11-14
Collate cert-manager presubmits into a single -bazel job

munnerz:


munnerz:
After this PR, the following presubmits will be enabled:

```
pull-cert-manager-bazel
pull-cert-manager-deps: takes a long time (and only required when Gopkg.* or vendor/ change)
pull-cert-manager-chart: requires docker
pull-cert-manager-e2e-v1-11: requires docker
pull-cert-manager-e2e-v1-12: requires docker
```

This removes:

```
pull-cert-manager-docs
pull-cert-manager-codegen
pull-cert-manager-unit
```

It also updates the `-chart` job to only run when `contrib/`, `hack/deploy` or `test/chart` change.

PR #129, opened 2019-01-09
Place ssh_key_secrets under decoration_config

munnerz:
/cc @wallrj


munnerz:
Er, that's weird.. you are a jetstack org member 🤔

PR #162, opened 2019-02-13
Install gcloud in bazelbuild base image

munnerz:
This is needed so we can use `gsutil` in our cert-manager release scripts.


munnerz:
cc @wallrj it looks like the katacoda-scenarios image build is failing due to npm not being installed/setup correctly after install.

Would you mind taking a look when you have a chance? 😄

PR #164, opened 2019-02-13
Bump cert-manager CI images

munnerz:


munnerz:
This is needed in order to make use of gsutil in our release scripts

PR #167, opened 2019-03-01
Bump test-infra version

munnerz:


munnerz:
This has now been deployed, so getting it merged.

If we encounter issues, we can revert this PR and redeploy Prow 😄

PR #178, opened 2019-03-26
Adds kube-oidc-proxy e2e test for Kubernetes v1.14

JoshVanL:
Signed-off-by: JoshVanL <vleeuwenjoshua@gmail.com>
/cc @simonswine 
/cc @munnerz 

Think we should keep 1.11 for the time being.. EKS is still lagging for example :roll_eyes:

### DIFF
--- a/config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
+++ b/config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
@@ -183,3 +183,54 @@ presubmits:
           type: Directory
     trigger: "(?m)^/test( e2e( v?1.13)?|)( \\[.+\\])?$"
     rerun_command: "/test e2e v1.13"
+
+  # kind based kube-oidc-proxy e2e job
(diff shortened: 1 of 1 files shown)


config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
```diff
@@ -183,3 +183,54 @@ presubmits:
           type: Directory
     trigger: "(?m)^/test( e2e( v?1.13)?|)( \\[.+\\])?$"
     rerun_command: "/test e2e v1.13"
+
+  # kind based kube-oidc-proxy e2e job
+  - name: pull-kube-oidc-proxy-e2e-v1-14
+    context: pull-kube-oidc-proxy-e2e-v1-14
+    # Match everything except PRs that only touch docs/
+    always_run: true
+    optional: true
```
simonswine: With this change we would run v1.13 and v1.14 for all PRs. I guess we only want to run one of them...

/assign @JoshVanL

PR #182, opened 2019-04-07
Update bazel image version

Julusian:


munnerz:
#183 encompasses this, and fixes up our stupid build system 😄

PR #183, opened 2019-04-08
Add image builder and pusher for CI images

munnerz:
This PR adds a small Golang program `builder` that can be used to build variants of Docker images defined in the jetstack/testing repo.

This PR moves us *away* from building the actual Docker images with Bazel, as after more learning about Bazel and how macros vs rules vs repository rules (workspace rules) work, I don't think there's a clean and efficient way to implement Dockerfile building without causing really long wait times during the Bazel analysis phase, or causing excessive IO juggling around the limitations of the analysis phase.

I've provided an example postsubmit job here too for Bazelbuild to switch over to use the new format.

Right now the builder uses Docker under the hood, but it should be possible to switch that out for Kaniko or even something like knative builds or google cloud build in future.

/cc @simonswine


config/jobs/testing/testing-postsubmits.yaml
```diff
@@ -0,0 +1,64 @@
...
+  env:
+  - name: GOOGLE_APPLICATION_CREDENTIALS
+    value: /creds/service-account.json
+  volumeMounts:
+    - name: creds
+      mountPath: /creds
+  volumes:
+  - name: creds
+    secret:
+      secretName: deployer-service-account
```
munnerz: This secret needs to be created & deployed still

images/builder/main.go
```diff
@@ -0,0 +1,317 @@
...
+	"log"
+	"math/rand"
+	"os"
+	"os/exec"
+	"path"
+	"strings"
+	"sync"
+	"time"
+
+	yaml "gopkg.in/yaml.v2"
```
simonswine: We should vendor this

images/builder/main.go
```diff
@@ -0,0 +1,317 @@
...
+	flag.StringVar(&buildDir, "build-dir", "", "path to a directory containing a build.yaml file")
+}
+
+func main() {
+	flag.Parse()
+
+	// validate flags
+	if errs := validateFlags(); len(errs) > 0 {
+		for _, err := range errs {
+			log.Printf(err.Error())
```
simonswine: `log.Print()`

WORKSPACE
```diff
@@ -139,101 +143,32 @@ yarn_install(
...
+#     name = "minikube-in-go",
+#     build_args = {
+#         "BAZEL_VERSION": "0.18.0",
+#         "GCLOUD_VERSION": "163.0.0",
+#     },
+#     build_matrix = {
+#         "KUBERNETES_VERSION": ["v1.9.6", "v1.8.10", "v1.7.15"],
+#     },
+#     dockerfile = "//legacy/images/minikube-in-go:Dockerfile",
+# )
```
simonswine: Remove comments?

config/jobs/testing/testing-postsubmits.yaml
simonswine: Happy to help with it

legacy/images/gcloud-in-go/BUILD.bazel
```diff
@@ -15,15 +15,15 @@ filegroup(
...
+# )
 
-docker_push(
-    name = "push",
-    bundle = ":gcloud-in-go",
-)
+# docker_push(
+#     name = "push",
+#     bundle = ":gcloud-in-go",
+# )
```
simonswine: Remove comments?

legacy/images/minikube-in-go/BUILD.bazel
```diff
@@ -15,17 +15,17 @@ filegroup(
...
+# )
 
-docker_push(
-    name = "push",
-    bundle = ":minikube-in-go",
-)
+# docker_push(
+#     name = "push",
+#     bundle = ":minikube-in-go",
+# )
```
simonswine: Remove comments?

config/jobs/testing/testing-presubmits.yaml
```diff
@@ -31,16 +31,14 @@ presubmits:
...
-    run_if_changed: "^(images/|legacy/images/).*$"
+  - name: pull-testing-verify-bazel
+    always_run: true
+    context: pull-testing-verify-bazel
     max_concurrency: 4
     agent: kubernetes
     decorate: true
     labels:
       preset-service-account: "true"
-      preset-dind-enabled: "true"
```
simonswine: Are we not using dind? Not too sure if like that.

Could we use long running dind instance (through stateful set maybe?) for enable docker caching

images/builder/main.go
munnerz: Bazel currently fetches it with a WORKSPACE rule, so we are pinned to a specific revision.

I opted not to actually vendor it into `vendor/` because that'd mean switching to actually use a dependency manager in this repo, and there are no other places that currently utilise anything outside of the Go stdlib.

WORKSPACE
munnerz: I'd commented these out in case we want to re-enable the Navigator jobs again in future, but happy to remove altogether for now and we can `git revert` in future if needs be?

config/jobs/testing/testing-presubmits.yaml
munnerz: The diff has rendered weirdly - as part of this PR, we'll no longer have a presubmit that builds images *at all*.

The rule in question that you're commenting on, is running `bazel build //...` (which is something we didn't previously run).

We can follow up with additional presubmits that automate testing images on each PR 😄

munnerz: > Could we use long running dind instance (through stateful set maybe?) for enable docker caching

Possibly, but this is also going to be super difficult to get set up and for now I just want to be able to push new images easily 😄

WORKSPACE
simonswine: I think it would be cleaner

config/jobs/testing/testing-presubmits.yaml
simonswine: Yeah makes more sense now with being aware no presubmit :slightly_smiling_face:

config/jobs/testing/testing-postsubmits.yaml
munnerz: Secret deployed

PR #196, opened 2019-04-08
Add tool for bumping Prow version in manifests

munnerz:
I'll then write another wrapper script that automatically creates a PR with these changes.

In future, we can include something here that actually overrides the `version` in prow/version with one detected automatically by running `gcloud container images list-tags`, allowing fully automated Prow image upgrades 😄


munnerz:
Closing in favour of an upcoming PR that will be a lot more magic 🙃

PR #202, opened 2019-04-09
Fixes kube-oidc-proxy-demo demo test

JoshVanL:
Signed-off-by: JoshVanL <vleeuwenjoshua@gmail.com>

/assign @simonswine 
/assign @munnerz

### DIFF
--- a/config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
+++ b/config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
@@ -42,9 +42,11 @@ presubmits:
     spec:
       containers:
       - image: golang:1.12.1
-        workingDir: /go/src/github.com/jetstack/kube-oidc-proxy/demo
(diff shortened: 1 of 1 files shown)


config/jobs/kube-oidc-proxy/kube-oidc-proxy-presubmits.yaml
```diff
@@ -42,9 +42,11 @@ presubmits:
     spec:
       containers:
       - image: golang:1.12.1
-        workingDir: /go/src/github.com/jetstack/kube-oidc-proxy/demo
+        workingDir: /go/src/github.com/jetstack/kube-oidc-proxy
```
munnerz: I don't think we should be specifying workingDir at all - afaik, it does *not* influence where your code is checked out to, and right now I think Prow is actually ignoring this option altogether.

JoshVanL: Sure, will remove all occurrences from the file if that's the case

PR #224, opened 2019-04-11
Tarmak disable bootstap py

simonswine:


config/jobs/tarmak/tarmak-presubmits.yaml
```diff
@@ -129,133 +79,7 @@
   trigger: "(?m)^/test( all| acceptance),?(\\s+|$)"
   rerun_command: "/test acceptance"
```
munnerz: I think the trigger and rerun command here should be removed, in order to force users of this template to set them explicitly?

simonswine: Makes sense

PR #229, opened 2019-04-23
Add Dockerfile for Terraform Google GKE cluster testing

wwwil:


images/terraform-google-gke-cluster/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+# limitations under the License.
+
+FROM debian:buster-slim
+
+RUN apt-get -y update \
+&& apt-get -y install --no-install-recommends \
+   ca-certificates \
+   curl \
+   unzip \
+&& curl https://releases.hashicorp.com/terraform/0.11.13/terraform_0.11.13_linux_amd64.zip -o terraform.zip \
```
munnerz: (may be best in a follow up): you can make the terraform version number be configurable via the `build.yaml` file, and then we'll be able to build a 'variant' of this image for each terraform version to make upgrading in the future easier to track.

PR #260, opened 2019-07-17
Add prometheus scrape annotations

simonswine:


munnerz:
We don't have Prow auto-deployment yet, so I'll apply/deploy these changes once I have a stable internet connection 😄 

/lgtm
/approve

PR #266, opened 2019-07-25
Remove ci-flightdeck-production-e2e

kragniz:


munnerz:
I've removed the token from the build cluster

/lgtm

PR #300, opened 2019-10-16
Update spyglass config for test coverage and testgrid

munnerz:

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -47,15 +47,34 @@ sinker:
 
 deck:
   spyglass:
-    size_limit: 500000000
+    size_limit: 500000000 # 500MB
     gcs_browser_prefix: https://console.cloud.google.com/storage/browser/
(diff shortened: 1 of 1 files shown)


config/config.yaml
```diff
@@ -47,15 +47,34 @@ sinker:
...
+    size_limit: 500000000 # 500MB
     gcs_browser_prefix: https://console.cloud.google.com/storage/browser/
-    viewers:
-      "started.json|finished.json":
-      - "metadata"
-      "build-log.txt":
-      - "buildlog"
-      "artifacts/junit.*\\.xml": # Remember to escape your '\' in yaml strings!
-      - "junit"
+    testgrid_config: gs://k8s-testgrid/config
```
munnerz: This file is public so I think it's fine to include here, and will allow spyglass to integrate with testgrid '''better''' (❓)

PR #310, opened 2019-11-12
Add presubmit for preflight

j-fuentes:
This adds a presubmit job for jetstack/preflight.

This one just runs the unit tests. If it works, I will continue adding more checks.

### DIFF
--- a/config/jobs/preflight/preflight-presubmits.yaml
+++ b/config/jobs/preflight/preflight-presubmits.yaml
@@ -1,0 +1,23 @@
+presubmits:
+  jetstack/preflight:
+
+  - name: pull-preflight-unit
+    cluster: gke
+    agent: kubernetes
+    decorate: true
+    always_run: true
+    max_concurrency: 8
+    spec:
+      containers:
+      - image: golang:1.12.7
+        args:
+        - make
+        - test
+        resources:
+          requests:
+            cpu: 500m
+            memory: 200Mi
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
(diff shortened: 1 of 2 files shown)


config/jobs/preflight/preflight-presubmits.yaml
```diff
@@ -0,0 +1,29 @@
...
+  jetstack/preflight:
+
+  - name: preflight-unit-tests
+    cluster: gke
+    agent: kubernetes
+    decorate: true
+    decoration_config:
+      ssh_key_secrets:
+      - jetstack-pulling-bot
+    clone_uri: "git@github.com:jetstack/preflight.git"
```
munnerz: You can remove this + the line above (as it won't work anyway as that secret doesn't exist in the gke build cluster). The repo will be public anyway 😄

config/jobs/preflight/preflight-presubmits.yaml
```diff
@@ -0,0 +1,29 @@
...
+    always_run: true
+    trigger: "(?m)^/test"
+    rerun_command: "/test"
```
munnerz: I think you can remove the `trigger` line, and then change `rerun_command` to `/test preflight-unit-tests`

j-fuentes: AFAIK `trigger` and `rerun_command` need to exist (I got an error before because `trigger` wasn't defined).

j-fuentes: > {"component":"checkconfig","error":"Either both of job.Trigger and job.RerunCommand must be set, wasnt the case for job \"preflight-unit-tests\"","file":"prow/cmd/checkconfig/main.go:180","func":"main.main","level":"fatal","msg":"Error loading Prow config.","time":"2019-11-12T18:08:46Z"}

exactly, `trigger` is also needed.

PR #317, opened 2019-12-02
Add hooks for release and build for preflight

j-fuentes:
In jetstack/preflight#11 we add these new make targets that are specially intended for prow.

Signed-off-by: Jose Fuentes <jsfuentescastillo@gmail.com>

### DIFF
--- a/config/jobs/preflight/preflight-postsubmits.yaml
+++ b/config/jobs/preflight/preflight-postsubmits.yaml
@@ -1,0 +1,108 @@
+presets:
+- labels:
+    preset-preflight-publish-bot-credentials: "true"
+  env:
+  - name: DOCKER_CONFIG
+    value: /etc/pusher-docker-config
+  volumes:
+  - name: pusher-config
+    secret:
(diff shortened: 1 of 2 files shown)


config/jobs/preflight/preflight-postsubmits.yaml
```diff
@@ -0,0 +1,108 @@
+presets:
+- labels:
+    preset-preflight-publish-bot-credentials: "true"
+  env:
+  - name: DOCKER_CONFIG
+    value: /etc/pusher-docker-config
+  volumes:
+  - name: pusher-config
+    secret:
+      secretName: preflight-publish-bot
```
j-fuentes: I am not sure how we handle these secrets. What can I do to create this secret if I want to push to quay.io?

munnerz: They need to be manually provisioned in our 'trusted' build cluster to do this, and a new robot account for preflight will need to be created.

PR #330, opened 2020-02-19
Add index periodical job for cert-manager/website

meyskens:
This adds a periodical job to update the search index of cert-manager.io

### DIFF
--- a/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
+++ b/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
@@ -26,8 +26,8 @@ periodics:
             key: ALGOLIA_WRITE_KEY
       resources:
         requests:
-          cpu: 2
-          memory: 4Gi
+          cpu: 1
+          memory: 2Gi
     dnsConfig:
       options:
         - name: ndots
--- a/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
+++ b/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
@@ -22,7 +22,7 @@ periodics:
       - name: ALGOLIA_WRITE_KEY
         valueFrom:
           secretKeyRef:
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
```diff
@@ -0,0 +1,34 @@
...
+      - scripts/index
+      env:
+      - name: ALGOLIA_WRITE_KEY
+        valueFrom:
+          secretKeyRef:
+            name: cert-mamager-website-algolia-key
+            key: ALGOLIA_WRITE_KEY
+      resources:
+        requests:
+          cpu: 2
```
munnerz: The VMs in this build cluster only have 2 CPUs, so probably best to set this to `cpu: 1` instead 😄 - I think you should also be able to reduce the memory request a bit? Perhaps `1Gi` or `2Gi`?

config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
```diff
@@ -0,0 +1,34 @@
...
+    containers:
+    - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:20191129-c49853e-1.13.4"
+      args:
+      - bash
+      - scripts/index
+      env:
+      - name: ALGOLIA_WRITE_KEY
+        valueFrom:
+          secretKeyRef:
+            name: cert-mamager-website-algolia-key
```
munnerz: manager not mamager 😄

meyskens: oops!

config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
meyskens: Done, would stay at 2 since we run node.js

PR #331, opened 2020-02-27
Enable label_sync for additional repos

munnerz:


munnerz:
This is how we configure new repos to have labels automatically setup/created/managed.

The actual labels that will be set can be seen here: https://github.com/jetstack/testing/blob/master/config/labels.yaml

It also handles migrating old labels to new labels too (e.g. renames)

cc @meyskens

PR #345, opened 2020-03-06
Add Kubernetes 1.11 and 1.17 e2e jobs

munnerz:
Let's see if we can get 1.11 to pass now that we're also aiming to support OpenShift 3 😄 

I've also marked the 1.17 presubmit as `always_run` but with `optional: true` - this will mean that all PRs will have results for this job, but won't require it to pass in order to merge. In a few days, we can disable the v1.15 job by default and mark the v1.17 job as required with minimal disruption to existing PRs 😄


munnerz:
Waiting for this to merge first: https://github.com/jetstack/cert-manager/pull/2664

/hold

PR #346, opened 2020-03-07
Run cert-manager periodic jobs every 2h

munnerz:


munnerz:
This should give us a more consistent set of results on testgrid, making it easier to cross-reference and generally get a 'signal' 😄

PR #350, opened 2020-03-09
Bump bazelbuild version used in testing repo

munnerz:


munnerz:
Tests will fail here until this is applied 🙈

PR #353, opened 2020-03-10
Set username and email argument when running transfigure

munnerz:
This should fix errors in the postsubmit job whereby it cannot detect the email address to use 😄


config/jobs/testing/testing-trusted.yaml
```diff
@@ -82,7 +82,9 @@ postsubmits:
         - ./config/jobs/ # path to prow job definitions
         - ./config/testgrid/dashboards.yaml # path to testgrid config
         - jetstack # name of the directory containing our testgrid configs
-        - jetstack/test-infra
```
munnerz: This format was invalid - it should just be the name of the repo, not the `org/repo` 😄

PR #357, opened 2020-03-13
Update cert-manager config for v0.15 & add new dashboards

munnerz:
Updates repository configuration for cert-manager v0.15.

I've also re-arranged config into `next` and `previous` directories, and created dashboards for us to monitor `previous`, `master` and `next` versions too.

I also fixed a few issues that I spotted as I was doing this 😄


config/config.yaml
```diff
@@ -48,7 +48,7 @@ branch-protection:
             - pull-cert-manager-bazel
             - pull-cert-manager-deps
             - pull-cert-manager-chart
-            - pull-cert-manager-e2e-v1-15
+            - pull-cert-manager-e2e-v1-17
```
wallrj: Is this Kubernetes 1.15 > 1.17 ?

munnerz: Yep

munnerz: We set `always_run: true` on the v1.17 job as well as the v1.15 job last week, so all recently opened PRs will already have test results for this job 😄 (gradual 'rollover' as it were). Now that we're seeing the job passing we're ready to switch it over and make it the required one (as we don't need to run both on every pull request!)

PR #380, opened 2020-05-04
Roll release-next for 0.15 release

meyskens:
This prepares the testing for release-previous to be 0.15 and release-next to be 0.16

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -48,7 +48,7 @@ branch-protection:
             - pull-cert-manager-bazel
             - pull-cert-manager-deps
             - pull-cert-manager-chart
-            - pull-cert-manager-e2e-v1-17
+            - pull-cert-manager-e2e-v1-18
         cert-manager-csi:
           protect: true
           required_status_checks:
@@ -252,29 +252,6 @@ tide:
   # Maintain separate cert-manager configuration to make it easy to enable code freeze
   - repos:
     - jetstack/cert-manager
-    excludedBranches:
-    - master
-    - release-0.15
-    labels:
-    - lgtm
-    - approved
-    - "dco-signoff: yes"
-    missingLabels:
-    - do-not-merge
-    - do-not-merge/blocked-paths
-    - do-not-merge/cherry-pick-not-approved
-    - do-not-merge/hold
-    - do-not-merge/invalid-owners-file
-    - do-not-merge/release-note-label-needed
-    - do-not-merge/work-in-progress
-    - needs-rebase
-    - needs-kind
-  - repos:
-    - jetstack/cert-manager
-    includedBranches:
-    - master
-    - release-0.15
-    milestone: v0.15
     labels:
     - lgtm
     - approved
--- a/config/jobs/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager-presubmits.yaml
@@ -60,7 +60,7 @@ presubmits:
     decorate: true
     branches:
     - master
-    - release-0.15
+    - release-0.16
     annotations:
       testgrid-create-test-group: 'true'
       testgrid-dashboards: jetstack-cert-manager-presubmits-blocking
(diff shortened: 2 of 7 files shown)


config/config.yaml
```diff
@@ -254,27 +254,7 @@ tide:
     - jetstack/cert-manager
     excludedBranches:
     - master
```
munnerz, line 256: We don't want to exclude any branches from this config else tide won't work on those branches. We only use `include/exclude` when code freeze is enabled in order to have different config per branch

config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
```diff
@@ -533,3 +533,60 @@ presubmits:
         options:
         - name: ndots
           value: "1"
+
+  - name: pull-cert-manager-e2e-v1-18
+    cluster: gke
+    context: pull-cert-manager-e2e-v1-18
+    always_run: true
```
munnerz, line 540: We don't want `always_run: true` along with `optional: true`

munnerz: Also: it looks like you didn't update this file by copying and pasting from `cert-manager-presubmits.yaml` in the parent directory as this same issue is still on the `v1-17` job? (I made the same mistake last time, which is why it's on that job and I think you've c+p'd that 😅)

config/config.yaml
meyskens: I see!

config/config.yaml
```diff
@@ -249,26 +249,6 @@ tide:
     - do-not-merge/release-note-label-needed
     - do-not-merge/work-in-progress
     - needs-rebase
-  # Maintain separate cert-manager configuration to make it easy to enable code freeze
```
munnerz, line 252: 👀 sorry I think we've got mixed up - this block *should* still exist, but you need to just remove the `excludedBranches` section from it as we have the same merge requirements for all branches in the repo when we aren't in code freeze

config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
```diff
@@ -537,8 +537,8 @@ presubmits:
   - name: pull-cert-manager-e2e-v1-18
     cluster: gke
     context: pull-cert-manager-e2e-v1-18
+    optional: false
```
munnerz, line 540: This is the 'previous' file - if we are switching the default Kubernetes e2e job from v1.17 to v1.18, we also need to update the branch protection config here: https://github.com/jetstack/testing/blob/a746b992b159beb6ddea486a3f383c060a4271d4/config/config.yaml#L36-L51

I think we'll want the same default across all branches come to think, unless we configure branch-specific merge protection config (which starts getting quite confusing..) 🙄

munnerz:
Also need to update milestone_applier config: https://github.com/jetstack/testing/blob/2dec3b891bc246b5050fb79f11bd24dca8d1b4df/config/plugins.yaml#L84-L101

config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -619,14 +619,14 @@ presubmits:
   - name: pull-cert-manager-e2e-v1-18
     cluster: gke
     context: pull-cert-manager-e2e-v1-18
-    always_run: false
-    optional: true
+    always_run: true
+    optional: false
```
munnerz, line 623: You need to set `optional: true` and `always_run: false` on the v1.17 job too :)

config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
```diff
@@ -372,7 +372,7 @@ presubmits:
     agent: kubernetes
```
munnerz, line 372: 2 lines up from here (https://github.com/jetstack/testing/pull/380/files#diff-8126d44f4ec7d37159faa0210dfb2e02R370) you've still got `always_run: true` on the v1.15 job from release-previous

PR #389, opened 2020-07-08
Add pull-cert-manager-e2e-v1-19 test

meyskens:
This adds pull-cert-manager-e2e-v1-19 to the presubmit ones to be manually triggered.
We should only add this test to the periodics once relese-next is v1.0.0 as 0.16 will not support 1.19


meyskens:
/hold

Due to a typo we were misinformed about removal of an API version. We can run all checks on Kubernetes 1.19

PR #392, opened 2020-07-23
Prepare for v0.16 and v1.0

meyskens:
/hold

This prepares the CI for the next release :tada:

### DIFF
--- a/config/jobs/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager-presubmits.yaml
@@ -60,7 +60,7 @@ presubmits:
     decorate: true
     branches:
     - master
-    - release-0.16
+    - release-1.0
     annotations:
       testgrid-create-test-group: 'true'
       testgrid-dashboards: jetstack-cert-manager-presubmits-blocking
@@ -96,7 +96,7 @@ presubmits:
     decorate: true
     branches:
     - master
-    - release-0.16
+    - release-1.0
     annotations:
       testgrid-create-test-group: 'true'
       testgrid-dashboards: jetstack-cert-manager-presubmits-blocking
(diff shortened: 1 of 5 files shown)


config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -60,7 +60,7 @@ presubmits:
     decorate: true
     branches:
     - master
-    - release-0.16
+    - release-1.0
```
wallrj, line 63: Does this configure prow to run this test before merging into either of these two branches?
Well then what about if we want to do a 0.16 patch release? It looks like we're configuring prow to no longer run these tests before merging stuff into the release-0.16 branch. Or have I got the wrong end of the stick?

config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
```diff
@@ -9,7 +9,7 @@ presubmits:
     agent: kubernetes
     decorate: true
     branches:
-    - release-0.15
+    - release-0.16
```
wallrj, line 12: Ah, I guess this answers my earlier question.

config/jobs/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
```diff
@@ -479,3 +479,61 @@ periodics:
       options:
       - name: ndots
         value: "1"
+
+- name: ci-cert-manager-previous-e2e-v1-18
```
wallrj, line 483: Does this mean we hadn't been periodically running e2e tests for CM 0.15 on k8s 1.18?
Was that just an oversight? Or is there a reason that 0.15 wasn't tested on k8s 1.18?

meyskens: we did not as on time of release k8s was not released, we did always test master with it

meyskens:
@wallrj for now it is edited by hand, a script would be nice but not always is the change this small

PR #398, opened 2020-09-30
Default config changes for Jetstack

zee-ahmed:
We're starting to roll out the adoption of Github and some of our repositories don't require Prow at Jetstack.

We want to make this optional and not required for every new and existing repository apart from those that require it today i.e. cert-manager, preflight, kube-oidc-proxy, etc.

I wasn't able to find much documentation for the changes I wanted to create but I'm hoping this should be OK going forward.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -215,18 +215,33 @@ tide:
   # Default tide config for all repos in the Jetstack org except cert-manager
   - orgs:
     - jetstack
-    - cert-manager
     excludedRepos:
     - jetstack/cert-manager
     - jetstack/tarmak
     - jetstack/cert-manager-csi
     labels:
     - lgtm
+    - approved
     missingLabels:
     - do-not-merge
     - do-not-merge/blocked-paths
     - do-not-merge/cherry-pick-not-approved
     - do-not-merge/hold
+    - do-not-merge/work-in-progress
+    - needs-rebase
+  # Default tide config for all repos in the cert-manager org
+  - orgs:
+    - cert-manager
+    labels:
+    - lgtm
+    - approved
+    - "dco-signoff: yes"
+    missingLabels:
+    - do-not-merge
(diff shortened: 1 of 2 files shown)


config/config.yaml
```diff
@@ -215,14 +215,26 @@ tide:
...
+    missingLabels:
+    - do-not-merge
+    - do-not-merge/blocked-paths
+    - do-not-merge/cherry-pick-not-approved
+    - do-not-merge/hold
+    - do-not-merge/work-in-progress
+    - needs-rebase
+  # Default tide config for all repos in the cert-manager org
+  - orgs:
+    - cert-manager
```
james-w, line 234: Will this work given that it's setting config for `jetstack` org projects?

meyskens: the one bot runs on both orgs, this copies the config over

meyskens:
@zee-ahmed LGTM when the CI is passing

config/config.yaml
```diff
@@ -236,6 +248,9 @@ tide:
   - repos:
     - jetstack/tarmak
     - jetstack/cert-manager-csi
+    - jetstack/preflight
```
meyskens, line 251: Should we double check this repo (testing) is still included?

zee-ahmed: Yes, we should! good spot.

zee-ahmed: This has now been added

config/config.yaml
james-w: Right, but what is the point of the `orgs` list? If it filters which orgs will be considered then I think this config section would do nothing, as it is saying to look at the `cert-manager` org for the projects `jetstack/cert-manager` etc.

meyskens: This line says to apply the "defaults" to all repos in the cert-manager org. jetstack/cert-manager is set as an exception here

PR #406, opened 2020-12-09
Remove OpenShift 3 tests

meyskens:
Forgot to remove these when removing legacy k8s


config/jobs/cert-manager/release-next/cert-manager-release-next-periodics.yaml
```diff
@@ -303,68 +303,4 @@ periodics:
...
-        type: Directory
-    - name: cgroup
-      hostPath:
-        path: /sys/fs/cgroup
-        type: Directory
-    dnsConfig:
-      options:
-      - name: ndots
-        value: "1"
+        value: "1"
```
jakexks, line 306:
```suggestion
        value: "1"

```

PR #408, opened 2020-12-23
Use devel/ci-run-e2e.sh instead of hack/ci/run-e2e-kind.sh

wallrj:
A followup to https://github.com/jetstack/testing/pull/407

```release-note
NONE
```


wallrj:
@meyskens @jakexks PTAL as the 1.20 e2e tests are all failing right now.

PR #429, opened 2021-02-12
Allow testing of cherrypicked 1.1 releases for cert-manager

jakexks:


simplyzee (approved):

config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
```diff
@@ -598,3 +598,61 @@ presubmits:
         options:
         - name: ndots
           value: "1"
+
+  - name: pull-cert-manager-e2e-v1-20
+    cluster: gke
+    context: pull-cert-manager-e2e-v1-20
+    optional: false
+    always_run: true
```
munnerz, line 606: We should also disable `always_run: true` for the 1.19 job if we are changing the default to be 1.20

PR #440, opened 2021-02-23
Move repo for cert-manager-webhook-example

jakexks:
For context: https://github.com/cert-manager/webhook-example/pull/23


config/jobs/cert-manager-webhook-example/cert-manager-webhook-example-presubmits.yaml
```diff
@@ -1,5 +1,5 @@
 presubmits:
-  jetstack/cert-manager-webhook-example:
```
munnerz, line 2: Can we move this to be a subdirectory of https://github.com/jetstack/testing/tree/master/config/jobs/cert-manager instead? We could then also look to collate cert-manager ones into its own subdirectory there too

jakexks: That seems sane

simplyzee (approved):

PR #444, opened 2021-03-05
Fix the name of the webhook-example repo

james-w:
```release-note
NONE
```


irbekrm (dismissed):
Looks good to me- I guess it would make sense to eventually also change the name [here](https://github.com/jetstack/testing/blob/master/config/config.yaml#L49) (and the name of the actual config).

jakexks (approved):
/lgtm

PR #447, opened 2021-03-15
Disable prow in some projects

j-fuentes:


j-fuentes:
This is now redundant as https://github.com/jetstack/testing/pull/448 disables prow for the jetstack organizations.

PR #471, opened 2021-04-09
Passes GCP creds file to crier

irbekrm:
Crier needs to push to GCS.

This PR gives it credentials of `jetstack-logs` GCP SA, stored in a manually created k8s secret. This is a temporary solution as we are going to re-build the Prow cluster.

```release-note
NONE
```


jakexks (approved):
/release-note-none
/approve
/lgtm

PR #475, opened 2021-04-12
Move release previous to be v1.3

irbekrm:
I am guessing that we only want current patches into `v1.3` cert-manager to run against Kubernetes `v1.20`? Same as we do with merges into master.
Currently we are trying to merge a patch into `release-1.3` and [it's pending](https://github.com/jetstack/cert-manager/pull/3886) - this PR attempts to fix that. 

/kind cleanup


jakexks:
/approve
/lgtm

I believe this should have been done right after releasing 1.3.0

jakexks (approved):

PR #481, opened 2021-04-27
Disable pre and post commit hooks for preflight as we are using Github actions now

j-fuentes:
We want to use prow just to dco checks, and not for building images or testing the code.


davidsbond (approved):

maelvls:
Not an expert in Prow configuration but I'd say this looks OK! 😅

/approve
/lgtm
/release-note none

PR #491, opened 2021-05-20
Require cert-manager presubmits only against k8s 1.21

irbekrm:
I think we only want to require 1.21 presubmits, at the moment 1.20 still shows up for new PRs, see https://github.com/jetstack/cert-manager/pull/4034


jakexks (approved):
/approve
/lgtm

PR #494, opened 2021-05-25
Temporarily revert "Require cert-manager presubmits only against k8s …

irbekrm:
…1.21"

This *temporarily* reverts the requirement for all PRs against `cert-manager` to pass the e2e test on k8s v1.21, so that we can get this cherry-pick in https://github.com/jetstack/cert-manager/pull/4058

After that we should re-instate the requirement.

This reverts commit 62f61d43b6f36af52a78276752d44c06c37d560d.


wallrj (approved):
/approve
/lgtm

PR #495, opened 2021-05-25
Revert "Temporarily revert "Require cert-manager presubmits only agai…

irbekrm:
…nst k8s 1.21""

This reverts commit 378c629a416014fc7b69cada480c7d2815b1b2e4.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -70,7 +70,7 @@ branch-protection:
             - pull-cert-manager-bazel
             - pull-cert-manager-deps
             - pull-cert-manager-chart
-            - pull-cert-manager-e2e-v1-20
+            - pull-cert-manager-e2e-v1-21
         cert-manager-csi:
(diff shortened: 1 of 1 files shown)


wallrj (approved):
/lgtm
/approve

But in future we could also configure per-branch branch-protection rules, so that we can have different  branch status check requirements for each release branch: https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/prow/cmd/branchprotector/README.md#scope

irbekrm:
> But in future we could also configure per-branch branch-protection rules, so that we can have different branch status check requirements for each release branch: https://github.com/kubernetes/test-infra/blob/f7e21a3c18f4f4bbc7ee170675ed53e4544a0632/prow/cmd/branchprotector/README.md#scope

#497

PR #498, opened 2021-05-26
Updating RBAC roles to match upstream kubernetes test-infra

simplyzee:
Due to recent changes of disabling ABAC. We now need to update our RBAC roles to ensure Prow and other services can run jobs successfully.

### DIFF
--- a/prow/cluster/sinker_rbac.yaml
+++ b/prow/cluster/sinker_rbac.yaml
@@ -5,39 +5,54 @@ metadata:
   name: "sinker"
 ---
 kind: Role
-apiVersion: rbac.authorization.k8s.io/v1beta1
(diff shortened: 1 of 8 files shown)


prow/cluster/sinker_rbac.yaml
```diff
@@ -5,39 +5,54 @@ metadata:
...
+    - delete
+    - list
+    - watch
+    - get
   - apiGroups:
-      - ""
+    - coordination.k8s.io
     resources:
-      - configmaps
+    - leases
```
irbekrm, line 25: I didn't even realize sinker has changed how it does leader election now.

irbekrm (approved):
Thank you very much for this!

PR #503, opened 2021-05-27
Request less cpu for e2e tests

irbekrm:
Else, we run out of capacity in Prow worker's cluster.

3500m should allow us to fit two jobs (2 pods) on a n-standard-8 node

### DIFF
--- a/config/jobs/cert-manager-csi/cert-manager-csi-presubmits.yaml
+++ b/config/jobs/cert-manager-csi/cert-manager-csi-presubmits.yaml
@@ -66,7 +66,7 @@ presubmits:
         - e2e
         resources:
           requests:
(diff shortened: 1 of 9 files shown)


config/jobs/cert-manager-csi/cert-manager-csi-presubmits.yaml
```diff
@@ -66,7 +66,7 @@ presubmits:
         - e2e
         resources:
           requests:
-            cpu: 6
+            cpu: 3500m
             memory: 12Gi
```
munnerz, line 70: Do the memory requests add up/are they proportional to the instance size and CPUs too?

munnerz (approved):
/lgtm

Issue #479, opened 2021-04-18
Fix failing Pod row on cert-manager-website-update-index


irbekrm:
Currently all [cert-manager-website-update-index](https://testgrid.k8s.io/jetstack-cert-manager-website#cert-manager-website-update-index) jobs are marked as failing due to the new TestGrid Pod row.

We fixed this issue for jobs running in the `build-infra-workers` cluster by enabling the GCS reporter on Prow's crier and manually creating some RBAC in the `build-infra-workers` cluster for crier so it the reporter can gather the necessary information- temp solution whilst we haven't yet automated the test infra setup.

`cert-manager-website-update-index` runs in our 'trusted' cluster, so perhaps it is a better solution to disable the Pod row on that particular job than hack together some temporary RBAC there.

The row can be disabled by setting `disable_prowjob_analysis` field on the particular test group in TestGrid config, like it's done [here](https://github.com/kubernetes/test-infra/blob/master/testgrid/cmd/configurator/prow.go#L133-L173).
However [Jetstack's TestGrid config](https://github.com/kubernetes/test-infra/blob/master/config/testgrids/jetstack/gen-config.yaml) is autogenerated using [transfigure](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure) using our [prow config](https://github.com/jetstack/testing/blob/master/config/config.yaml), job configs and a [base testgrid config](https://github.com/jetstack/testing/blob/master/config/testgrid/dashboards.yaml) and I was not able to find a way how to get this field generated using our current setup.

I have opened a [PR against kubernetes/test-infra](https://github.com/kubernetes/test-infra/pull/21842) that would allow doing this via a ProwJob annotation.

Alternative solution would be to create RBAC for Prow's crier in the `trusted` cluster.

irbekrm:
This was fixed in #508

PR #508, opened 2021-06-02
Disable prowjob analysis for jobs on trusted cluster

irbekrm:
See #479 for context.

This PR will update Jetstack's TestGrid config to disable 'Pod' row for jobs that run in 'trusted' cluster.
It will not itself modify the TestGrid config, but will instead trigger [this Prow job](https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-trusted.yaml#L63) which will regenerate the TestGrid config and create a PR against `k/test-infra`.
This will fix the issue with Prow jobs that run in 'trusted' cluster being marked as failed in TestGrid (see i.e https://testgrid.k8s.io/jetstack-cert-manager-website#cert-manager-website-update-index) by removing the pod row. Alternatively we could have provided 'trusted' cluster's creds to Prow- I chose to not do that for security reasons.

If you want to verify what TestGrid config gets generated, you can do so by copying the `/configurator` binary from `gcr.io/k8s-prow/transfigure` and running it with the configs in this PR `configurator --prow-config ./config/config.yaml --prow-job-config config/jobs/ --output-yaml --yaml config/testgrid/dashboards.yaml --oneshot --output generated.yaml` (I did that).

### DIFF
--- a/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
+++ b/config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
@@ -14,6 +14,7 @@ periodics:
   annotations:
     testgrid-dashboards: jetstack-cert-manager-website
     testgrid-alert-email: cert-manager-dev-alerts@googlegroups.com
+    testgrid-disable-prowjob-analysis: "true"
     description: Updates the algolia search index for the cert-manager website
   spec:
     containers:
--- a/config/jobs/testing/testing-trusted.yaml
+++ b/config/jobs/testing/testing-trusted.yaml
@@ -75,7 +75,7 @@ postsubmits:
       testgrid-disable-prowjob-analysis: "true"
     spec:
       containers:
-      - image: gcr.io/k8s-prow/transfigure
+      - image: gcr.io/k8s-prow/transfigure:v20210601-16a04c27e3
         command:
         - /transfigure.sh
         args:
--- a/config/jobs/testing/testing-trusted.yaml
+++ b/config/jobs/testing/testing-trusted.yaml
@@ -72,6 +72,7 @@ postsubmits:
     annotations:
       testgrid-dashboards: jetstack-testing-janitors
       testgrid-alert-email: cert-manager-dev-alerts@googlegroups.com
+      testgrid-disable-prowjob-analysis: "true"
     spec:
       containers:
       - image: gcr.io/k8s-prow/transfigure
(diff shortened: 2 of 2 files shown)


munnerz:
`by removing the pod row. Alternatively we could have provided 'trusted' cluster's creds to Prow`

I'm not against this change, but am confused by this statement ^ the Prow cluster _does_ have credentials to access the trusted cluster (these are required in order for the Pods to actually be created)

config/jobs/cert-manager/website/cert-manager-website-periodics.yaml
```diff
@@ -14,6 +14,7 @@ periodics:
   annotations:
     testgrid-dashboards: jetstack-cert-manager-website
     testgrid-alert-email: cert-manager-dev-alerts@googlegroups.com
+    testgrid-disable-prowjob-analysis: "true"
```
jakexks, line 17: This annotation's functionality exists after this PR: https://github.com/kubernetes/test-infra/pull/21842

munnerz: Does that mean we need to upgrade our Prow installation?

irbekrm: No, I think we generate TestGrid config [here](https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-trusted.yaml#L77) and that was already pulling the latest version of the `transfigure` container. As far as I understand nothing else should be looking at that annotation.
(I did look at Prow upgrade this am, but looks like there might be some larger changes (Plank being removed? https://github.com/kubernetes/test-infra/blob/master/prow/ANNOUNCEMENTS.md#breaking-changes) so maybe that's a separate PR)

irbekrm:
> `by removing the pod row. Alternatively we could have provided 'trusted' cluster's creds to Prow`
> 
> I'm not against this change, but am confused by this statement ^ the Prow cluster _does_ have credentials to access the trusted cluster (these are required in order for the Pods to actually be created)

I need to look again what it needs- I created #479 a while ago and seem to not have taken enough notes

jakexks (approved):
/approve

irbekrm:
jetstack-bot's PR to k/test-infra from this change looks ok https://github.com/kubernetes/test-infra/pull/22801/commits/dec99ffb081d6d4c1a19e28219ed377f0bf700ef

PR #509, opened 2021-06-03
Adds istio-csr istio testing version to v1.10.0 and bumps Kube to 1.21

JoshVanL:

/assign @jakexks 

```release-note
NONE
```

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -41,8 +41,10 @@ branch-protection:
             contexts:
             - dco
             - pull-istio-csr-verify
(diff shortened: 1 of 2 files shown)


irbekrm:
Looks good- I was wondering if it wouldn't make sense to keep the `v1.20` job, but have it not run always? In case there was ever a need to run the e2e tests against `v1.20`? Does istio-csr has a concept of supported Kubernetes releases?

irbekrm (approved):

PR #511, opened 2021-06-09
Explain why we went with 3500m for the cpu request, cf #510

maelvls:
I think we forgot to document why we changed the CPU value in #510.

```release-note
NONE
```


wallrj (approved):
Thanks for adding the note @maelvls 

/lgtm
/approve

PR #512, opened 2021-06-11
adding a periodic job for testing on eks cluster

RinkiyaKeDad:
Signed-off-by: RinkiyaKeDad <arshsharma461@gmail.com>

/assign @jakexks


config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -356,3 +356,52 @@ periodics:
...
         value: "1"
+
+- name: aws-tests
+  interval: 6h
+  agent: kubernetes
+  decorate: true
+  # extra_refs:
+  # - org: jetstack
+  #   repo: cert-manager
+  #   base_ref: master
```
RinkiyaKeDad, line 367: What do these do? Are they needed?

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -356,3 +356,52 @@ periodics:
...
+    testgrid-alert-email: cert-manager-dev-alerts@googlegroups.com
+    description: Runs the end-to-end test suite against a EKS cluster
+  # labels:
+  #   preset-service-account: "true"
+  #   preset-dind-enabled: "true"
+  #   preset-bazel-remote-cache-enabled: "true"
+  #   preset-bazel-scratch-dir: "true"
+  #   preset-cloudflare-credentials: "true"
+  #   preset-venafi-tpp-credentials: "true"
+  #   preset-venafi-cloud-credentials: "true"
```
RinkiyaKeDad, line 380: Which ones will we need?

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -356,3 +356,52 @@ periodics:
...
+  spec:
+    containers:
+    - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20210323-ad5071a-3.7.2
```
RinkiyaKeDad, line 383: How do I specify the `golang-aws` image that we built here?

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -356,3 +356,52 @@ periodics:
...
+        ./devel/run-e2e.sh --ginkgo.focus "Public ACME Server HTTP01 Issuer" --acme-server-url=https://acme-staging-v02.api.letsencrypt.org/directory --ingress-controller-domain=aws.e2e-tests.cert-manager.io --testing-acme-email=arshsharma461+acme-tests-1@gmail.com --kubernetes-config=/home/arsh/projects/forks/test-infra-1/aws/kubeconfig_cert-manager-cluster;
+      resources:
+        requests:
+          cpu: 3500m
+          memory: 12Gi
+      env:
+      - name: AWS_ACCESS_KEY_ID
+        value: "???"
+      - name: AWS_SECRET_ACCESS_KEY
+        value: "???"
```
RinkiyaKeDad, line 405: We can't have the access keys on GitHub so how will this work?

RinkiyaKeDad (commented):
Have a couple of doubts :)

config/jobs/cert-manager/cert-manager-periodics.yaml
irbekrm: It specifies what repository this Prow job clones when it runs.

config/jobs/cert-manager/cert-manager-periodics.yaml
irbekrm: You will first need to add the infra to build that image so it gets pushed to Jetstack Build Infra images GCR and then you can get the image sha from there. So probably the commits that build the image need to go in a PR before the PR that adds this Prow job.

config/jobs/cert-manager/cert-manager-periodics.yaml
irbekrm: It'll probably need to be a `Secret` in the cluster where the Prow job runs. The Prow job can then be configured to consume it via a preset like we do it with i.e Venafi creds- preset defined [here](https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/config.yaml#L22) and consumed added to the job [here](https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/cert-manager-periodics.yaml#L86). (The actual `Secret` would need to be created in the cluster separately).

config/jobs/cert-manager/cert-manager-periodics.yaml
RinkiyaKeDad: Thanks! Would you also know where it clones the repo? For example right now when we run the golang image we are in `/go` at the start. So if we specify this, would we be inside `/go/cert-manager` or would we still be in `/go` but have the `cert-manager` folder already cloned for us?

irbekrm: From a quick check it's cloned at `/home/prow/go/src/github.com/jetstack/cert-manager` and Prow sets that as `workingDir` for the test container.

RinkiyaKeDad: By `workingDir` do you mean that at the launch of the container we would be at `/home/prow/go/src/github.com/jetstack/cert-manager`? If yes, then if we add another repo would the first one be the `workingDir`?

irbekrm: > By workingDir do you mean that at the launch of the container we would be at /home/prow/go/src/github.com/jetstack/cert-manager? 

Yep, [the workingDir in the container spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.19/#container-v1-core).

> If yes, then if we add another repo would the first one be the workingDir

Looking at the code [here](https://github.com/kubernetes/test-infra/blob/cc3a0f99acfab2e07213e9a45937130c79bcc7fa/prow/pod-utils/decorate/podspec.go#L795-L803) looks like that would happen by default, but it seems like we can also [set `work_dir: true` for a particular ref](https://github.com/kubernetes/test-infra/blob/cc3a0f99acfab2e07213e9a45937130c79bcc7fa/prow/apis/prowjobs/v1/types.go#L907) to configure which one should become `workingDir`.
I've not tried cloning more than one myself.

RinkiyaKeDad (commented):
/hold because of things mentioned in the review.

jakexks (approved):

PR #518, opened 2021-06-29
Adds a new CI image with Bazel

irbekrm:
This PR builds a new image with Bazel, Docker, Node, Go for GitHub actions jobs that need to run Bazel.
This is mostly for the FIPS-compliant cert-manager building.

### DIFF
--- a/config/jobs/testing/testing-trusted.yaml
+++ b/config/jobs/testing/testing-trusted.yaml
@@ -195,6 +195,42 @@ postsubmits:
           capabilities:
             add: ["SYS_ADMIN"]
 
+  - name: post-testing-push-bazel-tools
+    cluster: trusted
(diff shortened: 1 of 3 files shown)


wallrj (approved):
I've read the changes @irbekrm 

But haven't tried to build the images or anything.
I can't tell where they'll be pushed nor how these and other jobs and images will eventually be used.

I'd recommend adding some comments in the jobs yaml file, the Dockerfile and elsewhere explaining how all these files are used....although I can see that the existing jobs don't have many comments either.

Or if you prefer, just `/unhold` this and improve things later.

/lgtm
/hold

irbekrm:
Thanks for the review Richard! 

Agree that the infra image building should be documented- I will do this separately in #520 since the process is the same for all of them.

/hold cancel

PR #521, opened 2021-07-02
Use a semver Docker tag

irbekrm:
Hardcode Docker tag for `bazel-tools` image for the specific variant built.
This change is needed because the Node version used is not a valid Docker tag.
It looks like hardcoding it is the simplest solution as the Node version cannot be specified by just a semver tag and parsing the NODE_VERSION in the script causes issues due to the special characters in it.

To try building the image (without pushing) run `bazel run //images/builder -- --build-dir $(pwd)/images/bazel-tools` from within this repo.

### DIFF
--- a/images/bazel-tools/build.yaml
+++ b/images/bazel-tools/build.yaml
@@ -5,6 +5,9 @@ variants:
     arguments:
       BASE_IMAGE: "eu.gcr.io/jetstack-build-infra-images/golang-dind@sha256:bf41f2a8f6191842ad3ff56a071448ed6a8bdcece4d6d62c5d91733c2f1f3541"
       NODE_VERSION: "10.24.0~dfsg-1~deb10u1"
+      # This DOCKER_TAG is the Docker tag that corresponds to the Node version
+      # we use. We don't use the Node version directly because it is not a valid
+      # Docker tag.
       DOCKER_TAG: "10.24.0"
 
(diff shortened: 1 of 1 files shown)


images/bazel-tools/build.yaml
```diff
@@ -5,8 +5,9 @@ variants:
     arguments:
       BASE_IMAGE: "eu.gcr.io/jetstack-build-infra-images/golang-dind@sha256:bf41f2a8f6191842ad3ff56a071448ed6a8bdcece4d6d62c5d91733c2f1f3541"
       NODE_VERSION: "10.24.0~dfsg-1~deb10u1"
+      DOCKER_TAG: "10.24.0"
```
maelvls, line 8: suggestion: let's remember why this DOCKER_TAG is used instead of just the Node version 👍

```suggestion
      # This DOCKER_TAG is the Docker tag that corresponds to the Node version
      # we use. We don't use the Node version directly because it is not a valid
      # Docker tag.
      DOCKER_TAG: "10.24.0"
```

maelvls (dismissed):
That looks good to me! Feel free to unhold

/lgtm
/hold

irbekrm:
Thank you @maelvls ! I've added the suggested change

irbekrm:
@maelvls it might need a GitHub style code approve 😅

maelvls (approved):

PR #523, opened 2021-07-06
Allows to trigger Venafi Cloud e2e tests separately

irbekrm:
This PR creates a new Prow job that can be run 'manually' with `/test pull-cert-manager-e2e-v1-21-feature-issuers-venafi-cloud` to run just the Venafi Cloud e2e tests.

This is useful as we sometimes disable these tests for regular presubmits/periodics due to flakiness- this way we can still run them manually from the CI when needed using the creds in the test cluster.


jakexks (approved):
/lgtm

PR #529, opened 2021-07-15
Make upgrade test run on each PR

irbekrm:
This PR makes the upgrade test to run on each pull request.

This is because our e2e tests are now running with experimental feature gates enabled and will also be running with the `gateway-shim` controller enabled and the Gateway API CRDs installed once [cert-manager#4158](https://github.com/jetstack/cert-manager/pull/4158) gets merged- so it seems to make sense to also run this as a smoke test that a simple installation still works.

The commit only has a few lines changed, but because I moved the job above two similar jobs, Git calculated the diff in a weird way.

### DIFF
--- a/config/jobs/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager-presubmits.yaml
@@ -574,6 +574,68 @@ presubmits:
         - name: ndots
           value: "1"
 
+  # Verifies upgrade from the latest published release with both Helm chart and
+  # static manifests. This is an optional test.
+  - name: pull-cert-manager-upgrade
+    # Run always
+    always_run: true 
+    optional: false 
+    # No more than 4 instances of this job at the same time.
+    max_concurrency: 4
+    # This job will run on Kubernetes cluster.
+    agent: kubernetes
+    # Pod utilities will be set up.
+    decorate: true
+    branches:
+    - master
+    - release-1.6
+    annotations:
+      description: Runs cert-manager upgrade from latest published release
+    labels:
+      preset-service-account: "true"
+      preset-dind-enabled: "true"
+      preset-bazel-remote-cache-enabled: "true"
+      preset-bazel-scratch-dir: "true"
+    spec:
+      containers:
+      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20210831-4cf7b0b-4.2.1
+        args:
+        - runner
+        - make
+        - cluster
+        - verify_upgrade
+        resources:
+          requests:
+            cpu: 3500m
+            memory: 12Gi
+        env:
+        # Used by https://github.com/jetstack/cert-manager/blob/master/devel/cluster/create-kind.sh
+        - name: K8S_VERSION
+          value: "1.22"
+        securityContext:
+          privileged: true
+          capabilities:
+            add: ["SYS_ADMIN"]
+        volumeMounts:
+        - mountPath: /lib/modules
+          name: modules
+          readOnly: true
+        - mountPath: /sys/fs/cgroup
+          name: cgroup
+      volumes:
+      - name: modules
+        hostPath:
+          path: /lib/modules
+          type: Directory
+      - name: cgroup
+        hostPath:
+          path: /sys/fs/cgroup
+          type: Directory
+      dnsConfig:
+        options:
+        - name: ndots
+          value: "1"
+
   # An E2E test job to allow us to manually trigger the Venafi TPP E2E tests
   # with the following GitHub comment:
   #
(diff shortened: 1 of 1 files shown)


maelvls:
Might it be worth documenting this change in `content/en/docs/contributing/release-process.md`? There is a table that details which tests are running on every PR/periodically. It looks like this:


| Release branch | Prow configuration            | Dashboard                 | Kubernetes versions tested   |  Periodicity  |
| :------------: | :---------------------------- | :------------------------ | :--------------------------- | :-----------: |
|      PRs       | [`presubmits.yaml`][]         | [`presubmits-blocking`][] | 1.21                         |  On each PR   |
|     master     | [`periodics.yaml`][]          | [`master`][]              | 1.16, 1.17, 1.18, 1.19, 1.21 | Every 2 hours |
|  release-1.5   | [`next-periodics.yaml`][]     | [`next`][]                | 1.16, 1.17, 1.18, 1.19, 1.21 | Every 2 hours |
|  release-1.4   | [`previous-periodics.yaml`][] | [`previous`][]            | 1.16, 1.17, 1.18, 1.19, 1.21 | Every 2 hours |
|  release-1.3   | N/A                           |                           | N/A                          |      N/A      |

[`presubmits.yaml`]: https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/cert-manager-presubmits.yaml
[`periodics.yaml`]: https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/cert-manager-periodics.yaml
[`next-periodics.yaml`]: https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/release-next/cert-manager-release-next-periodics.yaml
[`previous-periodics.yaml`]: https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
[`presubmits-blocking`]: https://testgrid.k8s.io/jetstack-cert-manager-presubmits-blocking
[`master`]: https://testgrid.k8s.io/jetstack-cert-manager-master
[`next`]: https://testgrid.k8s.io/jetstack-cert-manager-next
[`previous`]: https://testgrid.k8s.io/jetstack-cert-manager-previous

config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -524,47 +524,44 @@ presubmits:
...
           value: "1"
 
-  # An E2E test job to allow us to manually trigger the Venafi TPP E2E tests
-  # with the following GitHub comment:
-  #
-  #  /test pull-cert-manager-e2e-v1-21-feature-issuers-venafi-tpp
-  #
-  # See https://github.com/jetstack/cert-manager/issues/3555
-  #
-  - name: pull-cert-manager-e2e-v1-21-feature-issuers-venafi-tpp
```
maelvls, line 534: question: why do we remove `pull-cert-manager-e2e-v1-21-feature-issuers-venafi-tpp`?

config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -524,47 +524,44 @@ presubmits:
...
-    always_run: false
-    optional: true
+  # Verifies upgrade from the latest published release with both Helm chart and
+  # static manifests. This is an optional test.
+  - name: pull-cert-manager-upgrade-v1-21
+    # Run only when requested.
+    always_run: true
+    optional: false
```
maelvls, line 532: It says "run only when requested", I guess the comment does not match `optional: false`?

config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -524,47 +524,44 @@ presubmits:
...
-        - devel/ci-run-e2e.sh
-        - -ginkgo.focus
-        - '\[Feature:Issuers:Venafi:TPP\]'
+        - make
+        - cluster
+        - verify_upgrade
         resources:
           requests:
             cpu: 3500m
             memory: 12Gi
```
maelvls, line 560: note: I didn't realize our e2e needed so much memory 🤯

maelvls:
/lgtm
/hold

(feel free to unhold since the PR comments I added are just nits)

config/jobs/cert-manager/cert-manager-presubmits.yaml
irbekrm: I think we don't strictly need that much (or maybe we do, I haven't checked how much we consume).
We use n1-standard-8 machines for workers- the 3500m cpu means 2 jobs can fit on a node which also has 32 GB of memory so allocating 12GB to each kind of makes sense, as you wouldn't fit more jobs onto the same node anyway.

Maybe worth exploring if we could use machines with more CPU vs memory.

config/jobs/cert-manager/cert-manager-presubmits.yaml
irbekrm: Thanks, that's a good catch!

irbekrm:
/hold

The logic how we calculate last published release is wrong, see https://github.com/jetstack/cert-manager/pull/4218#issuecomment-884089497

wallrj (dismissed):
This seems useful @irbekrm 
but needs rebasing
and maybe updating in preparation for the cert-manager 1.6 release.

wallrj (approved):
Thanks Irbe.

/lgtm
/approve

PR #530, opened 2021-07-16
Adds cert-manager/trust CI tests

JoshVanL:

```release-note
NONE
```

### DIFF
--- a/config/jobs/cert-manager-trust/OWNERS
+++ b/config/jobs/cert-manager-trust/OWNERS
@@ -1,0 +1,6 @@
+approvers:
+- joshvanl
+reviewers:
+- joshvanl
+labels:
+- area/cert-manager-trust
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -28,6 +28,7 @@ triggers:
   - jetstack/cert-manager-csi
   - cert-manager/istio-csr
   - cert-manager/policy-approver
+  - cert-manager/trust
   trusted_org: cert-manager
   only_org_members: true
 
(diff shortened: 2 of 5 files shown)


config/plugins.yaml
```diff
@@ -133,6 +138,8 @@ milestone_applier:
     master: v0.0
   cert-manager/policy-approver:
     master: v0.0
+  cert-manager/trust:
+    master: v0.0
```
munnerz, line 142: hmm.. is `v0.0` correct here? I see it for istio-csr & policy-approver too, but that's surely never a milestone we'd use?

JoshVanL: No, that is intended as a placeholder but istio-csr at least hasn't been updated :sweat_smile: 

Should this be removed completely?

config/jobs/cert-manager-trust/OWNERS
```diff
@@ -0,0 +1,6 @@
+approvers:
+- joshvanl
+reviewers:
+- joshvanl
+labels:
+- area/cert-manager-trust
```
SgtCoDFish, line 6: suggestion: I lose track of what prow actually cares about so I don't know how much `OWNERS` matters, but if we're adding it, should it at least contain the current members of the cert-manager-maintainers group as approvers and reviewers?

I note the reference to "milestone-maintainers" in `config/plugins.yaml` in this PR

SgtCoDFish (commented):
minor question from me

This PR generally looks totally unobjectionable to me and we can always change it later, but then I won't pretend to know much about prow's mysticism :grin:

config/jobs/cert-manager-trust/OWNERS
JoshVanL: My understanding is that this will automatically ping me to review if a PR is opened to change anything in this dir.

Happy to change it but don’t think it matters too much… a testing PR is normally always followed up by a slack message.

SgtCoDFish (approved):
/lgtm

PR #537, opened 2021-07-30
Updates our kube v1.22 version to V1.22.0-rc.0

JoshVanL:

Hopefully this is the only thing I need to do :smile: 

/assign @irbekrm 

```release-note
NONE
```


jakexks (approved):
/lgtm
I think this is right, plus a follow-up to create-kind.sh in cert-manager

PR #540, opened 2021-08-05
Bump k8s version for v1.22 image

irbekrm:
Bump k8s version to v1.22.0

Once this image gets built and we have  seen e2e tests pass with that, I'd like to make this run by default on presubmits.


wallrj (approved):
/lgtm

PR #543, opened 2021-08-06
Enable Gateway API Feature gate on k8s 1.19+

jakexks:
This will allow the tests in https://github.com/jetstack/cert-manager/pull/4310 to pass.

/release-note-none

### DIFF
--- a/config/jobs/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager-periodics.yaml
@@ -286,9 +286,9 @@ periodics:
       env:
       - name: K8S_VERSION
         value: "1.19"
-      # Enable CertificateSigningRequest e2e tests for v1.19+ clusters
+      # Enable CertificateSigningRequest and Gateway API e2e tests for v1.19+ clusters
       - name: FEATURE_GATES
-        value: "ExperimentalCertificateSigningRequestControllers=true"
+        value: "ExperimentalCertificateSigningRequestControllers=true\\,ExperimentalGatewayAPISupport=true"
       securityContext:
         privileged: true
         capabilities:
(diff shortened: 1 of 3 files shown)


SgtCoDFish (dismissed):
/lgtm

Based on [this comment](https://github.com/jetstack/cert-manager/pull/4310#issuecomment-893494779) the change looks correct, although I've not tested this locally or anything.

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -286,9 +286,9 @@ periodics:
       env:
       - name: K8S_VERSION
         value: "1.19"
-      # Enable CertificateSigningRequest e2e tests for v1.19+ clusters
+      # Enable CertificateSigningRequest and Gateway API e2e tests for v1.19+ clusters
       - name: FEATURE_GATES
-        value: "ExperimentalCertificateSigningRequestControllers=true"
+        value: "ExperimentalCertificateSigningRequestControllers=true\\,ExperimentalGatewayAPISupport=true"
```
irbekrm, line 291: Why do we need double backslash?

SgtCoDFish:
/hold


holding out of caution based on [this comment](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1628252005166600?thread_ts=1628251701.164900&cid=CDEQJ0Q8M) in slack; if we're happy that it won't be an issue, we can unhold

jakexks:
Feature gates have to exist or the controller just exits. I think this means it should only be on release-next, not on any other tests.

config/jobs/cert-manager/cert-manager-periodics.yaml
jakexks: As helm interprets a comma as a list of values rather than a single string containing a comma, it needs to be passed to helm as `\,`. In YAML to get a literal backslash you should use `\\`

jakexks:
Updated to only update master and release-next (which is currently release-1.5)

SgtCoDFish (approved):
/lgtm

PR #544, opened 2021-08-06
\\ -> \\\\

jakexks:
Through a combination of YAML and bash the escaped backslashes weren't getting through.


SgtCoDFish (approved):
/lgtm

PR #545, opened 2021-08-06
\\\\, -> ,

inteon:
\\\\\\\\, causes some problems with the tests; lets instead fix the scripts in cert-manager and pass the string correctly here


inteon:
\release-note-none

jakexks (approved):
/lgtm

PR #551, opened 2021-08-11
destroying aws infra after tests

RinkiyaKeDad:
Signed-off-by: Arsh Sharma <arshsharma461@gmail.com>

/assgin @jakexks @irbekrm

### DIFF
--- a/config/jobs/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager-periodics.yaml
@@ -526,7 +526,8 @@ periodics:
         cd .. && \
         cd .. && \
         cd cert-manager && \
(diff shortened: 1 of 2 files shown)


config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -521,13 +521,14 @@ periodics:
         cd test-infra/aws && \
         terraform init && \
         terraform apply -auto-approve && \
+        trap 'terraform destroy -auto-approve' ERR && \
```
irbekrm, line 524: Suggestion: move the trap to line 522 - I think it needs to be defined before running `terraform apply` otherwise it won't run if `terraform apply` exits. (As I understand it works similarly to `defer` statement in Go if that helps).

irbekrm (changes requested):
Looks good to me, I've added one small suggestion to move `trap` to a line above, otherwise happy to lgtm the PR.
Thank you for the contribution, glad we'll have these tests now :)

RinkiyaKeDad:
Thanks for reviewing @irbekrm :) I've rearranged the commands, PTAL.

irbekrm (approved):

PR #552, opened 2021-08-11
Export GINKGO_SKIP/GINKGO_FOCUS env vars in test containers

irbekrm:
This PR:

- defines new presets where env vars can be defined to be exported in tests

- all tests running against Kubernetes v1.19 and newer will have `FETURE_GATES` set to `ExperimentalCertificateSigningRequestControllers=true`

- all tests running against Kubernetes v1.18 and older will have `FEATURE_GATES` set to `ExperimentalCertificateSigningRequestControllers=false`

- hopefully fixes the [currently red aws tests](https://testgrid.k8s.io/jetstack-cert-manager-master#aws-tests) by ensuring the Gateway tests are skipped

- all tests will skip Venafi Cloud e2e tests and a new periodic is added that will run the Venafi Cloud e2e tests only every 24 hours.

### DIFF
--- a/config/jobs/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager-periodics.yaml
@@ -84,7 +84,7 @@ periodics:
     preset-bazel-scratch-dir: "true"
     preset-cloudflare-credentials: "true"
     preset-venafi-tpp-credentials: "true"
-    preset-venafi-cloud-credentials: "true"
+    preset-k8s-pre-v19-env: "true"
   annotations:
     testgrid-create-test-group: 'true'
     testgrid-dashboards: jetstack-cert-manager-master
@@ -103,9 +103,6 @@ periodics:
       env:
       - name: K8S_VERSION
         value: "1.16"
-      # Disable CertificateSigningRequest e2e tests for pre v1.19 clusters
-      - name: FEATURE_GATES
-        value: "ExperimentalCertificateSigningRequestControllers=false"
       securityContext:
         privileged: true
         capabilities:
@@ -150,7 +147,7 @@ periodics:
     preset-bazel-scratch-dir: "true"
     preset-cloudflare-credentials: "true"
     preset-venafi-tpp-credentials: "true"
-    preset-venafi-cloud-credentials: "true"
+    preset-k8s-pre-v19-env: "true"
   spec:
     containers:
     - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20210323-ad5071a-3.7.2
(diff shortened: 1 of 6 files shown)


JoshVanL:
lgtm @irbekrm, should we also include the env vars for the CSR experiment feature gate?

irbekrm:
@JoshVanL yeah I think that'll make sense since the presets are for split on v1.19 Kubernetes- I'll change add that, thanks

irbekrm:
Thanks @JoshVanL I've updated the PR.

irbekrm:
Once this PR gets merged, it should be possible to successfully run all the tests (Venafi Cloud that are currently failing should be skipped) for [cert-manager#4338](https://github.com/jetstack/cert-manager/pull/4338) and merge that PR. 

After [cert-manager#4338](https://github.com/jetstack/cert-manager/pull/4338) too gets merged, the cert-manager tests should be unblocked.

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -272,7 +263,7 @@ periodics:
     preset-bazel-scratch-dir: "true"
     preset-cloudflare-credentials: "true"
     preset-venafi-tpp-credentials: "true"
-    preset-venafi-cloud-credentials: "true"
+    preset-k8s-v19-plus-env: "true"
```
munnerz, line 266: Bit of a nit, but it took me quite a while to actually work out what this preset does due to the name of it being a description of the environment rather than a description of what it does (e.g. `preset-enable-k8s-csr-api-tests`). Could the preset be renamed?

Also, it may be better to _invert_ this so that eventually it is not required once k8s <1.19 is no longer in use? So.. `preset-skip-k8s-csr-tests` added to each e2e job that is LESS THAN kubernetes 1.19?

wallrj (approved):
Thanks @irbekrm 

I think I get the gist of this. 
Nice that the FOCUS and SKIP configuration can be applied as presets to multiple jobs.

Let's merge and see if this works.

/lgtm
/approve

config/jobs/cert-manager/cert-manager-periodics.yaml
munnerz: Ah, agreed for _gateway_ - but I was referring to the `csr` (i.e. kubernetes CertificateSigningRequest) feature gates here, rather than gateway... perhaps I got mixed up :)

irbekrm: Ahh no that Gateway bit didn't make any sense- the preset is used to enable/disable both feature gates https://github.com/jetstack/testing/commit/875eb40cef7f9b08f1b74b1f78da0780f5f7e8ec#diff-ce3d91fd77957f12a21c34a9d5b2805cf9c1b0fd5be4cc57a6372e9bc7fe500eL102 so we can't really split I think, IMO it's still better to be explicit that those both are enabled by default for as long as they are feature gates.

PR #555, opened 2021-08-12
fixing path in aws periodic job

RinkiyaKeDad:
Signed-off-by: RinkiyaKeDad <arshsharma461@gmail.com>

The periodic job apparently has a different path from the presubmit one.

Periodic job:
image

Presubmit job:
image

/release-note-none

### DIFF
--- a/config/jobs/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager-periodics.yaml
@@ -497,6 +497,10 @@ periodics:
   interval: 48h
   agent: kubernetes
   decorate: true
+  extra_refs:            # Periodic job doesn't clone any repo by default, needs to be added explicitly
+    - org: cert-manager
+      repo: test-infra
+      base_ref: master
   annotations:
     testgrid-create-test-group: 'true'
     testgrid-dashboards: jetstack-cert-manager-master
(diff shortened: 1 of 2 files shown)


irbekrm:
Thanks for the PR @RinkiyaKeDad.

> The periodic job apparently has a different path from the presubmit one.

I think the difference is that [presubmit by default clones the Git repository for which the presubmit is configured, whereas a periodic does not](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs).
So I think it might be the case that in the aws presubmit the repo gets cloned twice - once by Prow and once when you run `git clone` and if Prow _does_ clone it it puts it at `/home/prow/go/src/github.com/..`.

Do you think it would make sense to allow Prow to clone both repos by [configuring `extra_refs` on job spec](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs) instead of running `git clone` in the script? That would then be similar to [how we do that for other jobs](https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/cert-manager-periodics.yaml#L76)?

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -497,6 +497,10 @@ periodics:
   interval: 48h
   agent: kubernetes
   decorate: true
+  extra_refs:            # Periodic job doesn't clone any repo by default, needs to be added explicitly
```
irbekrm, line 500: Thanks for the comment 🙏🏼

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -497,6 +497,10 @@ periodics:
...
+    - org: cert-manager
+      repo: test-infra
+      base_ref: master
```
irbekrm, line 503: Could we also clone the cert-manager repo in the same way? (Both here and in the presubmit job.)

config/jobs/cert-manager/cert-manager-periodics.yaml
```diff
@@ -516,19 +520,20 @@ periodics:
       - -c
       - |
         set -euo && \
-        git clone https://github.com/jetstack/cert-manager.git && \
-        git clone https://github.com/cert-manager/test-infra.git && \
-        cd test-infra/aws && \
+        ls && \
+        pwd && \
+        cd /home/prow/go/src/github.com/jetstack/cert-manager/test-infra/aws && \
```
irbekrm, line 525: I _think_ the path might be `/home/prow/go/src/github.com/cert-manager/test-infra/aws`? (i.e `/home/prow/go/src/github.com/<org>/<repo>/aws`
Same on line 536 and 812, 823 in the other file.

irbekrm (commented):
Thanks for updating the PR Arsh!

I've left some more comments- I think it would make sense to clone both cert-manager and test-infra repos with extra refs, let me know if you agree.

Otherwise happy to merge this, once that's done we can exec to the pod and verify what the paths are.

config/jobs/cert-manager/cert-manager-periodics.yaml
RinkiyaKeDad: This was the initial path so I added that but yeah I'm pretty sure we'll need a few more PRs to figure out where stuff is :)

irbekrm:
>Invalid
job aws-tests-pre on repo jetstack/cert-manager: the following refs
specified more than once:

I think this is my bad, I said we should clone cert-manager both for presubmit and periodic, but in fact Prow already clones it for the presubmit.
So, I guess if you remove 
```
    - org: jetstack
      repo: cert-manager
      base_ref: maste
``` 
from the presubmit it should hopefully work 😅

irbekrm:
I've tested running [pull-testing-config](https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-presubmits.yaml#L5) locally and it looks like deleting [the extra ref for cert-manager from the presubmit job](https://github.com/RinkiyaKeDad/testing/blob/aws_periodical_job_fix/config/jobs/cert-manager/cert-manager-presubmits.yaml#L789-L791) fixes the error.

(To try it out `docker run -it -v $(pwd):/config  gcr.io/k8s-prow/checkconfig:v20210409-985ef5e721 -config-path=/config/config/config.yaml -strict=true -job-config-path=/config/config/jobs -plugin-config=/config/config/plugins.yaml` from the root of this repo)

RinkiyaKeDad:
Thanks a lot @irbekrm! The tests are passing now, please approve :D

irbekrm (approved):

PR #558, opened 2021-08-13
Enable Gateway API features in release-previous

wallrj:
* Enable Gateway API features in K8S >=1.19
* Explicitly disable it in K8S <1.19, so that if the feature graduates and is turned on by default, it doesn't start failing on old unsupported cluster versions
* Added a missing 1.18 job to the release-previous pre-submits.

This will fix the E2E failures on K8S >=1.19 seen here: https://testgrid.k8s.io/jetstack-cert-manager-previous

```release-note
Enable Gateway API features in release-previous
```


SgtCoDFish (approved):
/lgtm
/approve

PR #562, opened 2021-08-26
cert-manager release: use make rather than calling a script directly

SgtCoDFish:
This is possible now that https://github.com/cert-manager/release/pull/41 has been merged; it allows us to have the makefile as a one-stop-shop for all commands we might need when interacting with the project


SgtCoDFish:
@JoshVanL it looks like this might need you to do a PR review + approve, too :cry:

JoshVanL (approved):

PR #567, opened 2021-09-06
Use label sync on cert-manager/policy-approver and remove from job list

JoshVanL:

I have gone with how `cert-manager/csi-lib` has been setup which should hopefully work much better.

/assign @irbekrm 

```release-note
NONE
```


irbekrm:
We should probably at some point document the label sync approach vs explicitly adding the jobs to the config files.

/lgtm
/approve

irbekrm (approved):

PR #574, opened 2021-09-07
Bump prow

irbekrm:
This PR updates all Prow components to the latest version, passes some component flags that are now required and updates the `ProwJob` CRD.

This change has already been applied to the cluster and seems to be working.


irbekrm:
/release-note-none

(We should probably remove the release note requirement for this repo)

PR #578, opened 2021-09-08
Fix invocation of make in cert-manager/release presubmit

SgtCoDFish:
```release-note
NONE
```


SgtCoDFish:
@JoshVanL I think we need an approving review again :sweat_smile:

JoshVanL:
> @JoshVanL I think we need an approving review again

urgh.

PR #585, opened 2021-09-27
Bazel version bump

irbekrm:
This PR:

- bumps Bazel version `v3.7.2` -> `v4.2.1` for cert-manager tests (we were already running the Bazel-experimental job with this version, so it should work)
- bumps Bazel version for jobs against this repo `v2.2.0` -> `v4.2.1` (I have this version locally and can run tests against this repo, so it should work)
- temporarily removes 'experimental' Bazel jobs for cert-manager `master` and `release-1.16` branches. There is currently no newer version of Bazel than `v4.2.1` and there is no point to run the same test twice against the same version. This job should be re-instated once Bazel `v5.0.0` is released and we have an image for that https://github.com/bazelbuild/bazel/releases

### DIFF
--- a/config/jobs/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager-periodics.yaml
@@ -34,39 +34,8 @@ periodics:
         - name: ndots
           value: "1"
 
-- name: ci-cert-manager-bazel-experimental
-  interval: 2h
-  agent: kubernetes
-  decorate: true
-  extra_refs:
-  - org: jetstack
-    repo: cert-manager
-    base_ref: master
-  labels:
-    preset-service-account: "true"
-    preset-bazel-scratch-dir: "true"
-  annotations:
-    testgrid-create-test-group: 'true'
-    testgrid-dashboards: jetstack-cert-manager-master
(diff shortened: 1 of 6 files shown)


SgtCoDFish:
question: I checked out these changes locally and there are still several references to the old bazel version in `config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml` and `config/jobs/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml`.

Is that intentional, or should they be updated too?

irbekrm:
Thanks @SgtCoDFish - it is intentional, for the previous release the minimum required Bazel version to build it was v3.5.0 and we built the v1.5.3 release with Bazel v3.5.0, so I though it makes sense to keep the 3.X version.

 I'm bumping it for release-next/master because I want to bump rules_go as part of upgrading to Go 1.17 and upgrading some/libs dependencies in `cert-manager` and latest rules_go requires Bazel v4.X https://github.com/bazelbuild/rules_go/releases/tag/v0.28.0

SgtCoDFish (approved):
/lgtm

Thanks for explaining! In that case, looks good :grin:

PR #590, opened 2021-10-04
Bump bazel-tools image

irbekrm:
Bumps version of bazelbuild image used as a base for for bazel-tools so that a newer version gets built that has Go 1.17 and Bazel 1.4.2

Also change how the image is tagged to include the version of Go and Bazel as we care about those most.


irbekrm:
/hold

forgot to bump the actual sha

images/bazel-tools/build.yaml
```diff
@@ -3,14 +3,18 @@ name: bazel-tools # Name of the image to be built
 variants:
   "10.24":
     arguments:
-      BASE_IMAGE: "eu.gcr.io/jetstack-build-infra-images/golang-dind@sha256:bf41f2a8f6191842ad3ff56a071448ed6a8bdcece4d6d62c5d91733c2f1f3541"
+      BASE_IMAGE: "eu.gcr.io/jetstack-build-infra-images/golang-dind@sha256:a71c5e6e8cb6875e9f6a43b722254693759b5db77566e25ba9eb52ca74b92e08"
+      # Version of Bazel that is bundled in the BASE_IMAGE
+      BAZEL_VERSION: "1.4.2"
```
wallrj, line 8: Should this be 4.2.1 ?

wallrj (changes requested):
I think the version of Bazel is wrong. And if I'm right, how come none of the tests failed? Seems like it might be too easy for this version to get out of sync with the actual version of Bazel embedded in the Docker image. Or have I misunderstood?

images/bazel-tools/build.yaml
irbekrm: Thanks for catching this Richard- it's a typo, the image does have the right version, I've just updated it

Yeah agree that it _can_  easily get incorrect (as we just saw), but I still would rather have the version number embedded in the tag. I think it's ok for whoever is adding this image tag to some code to still have to double check that everything works, but then at least the reader of the code doesn't have guess what versions of things are being used.

irbekrm: Unfortunately we can't really check that the tags are correct since we use the same build script to make these images

images/bazel-tools/build.yaml
```diff
@@ -3,14 +3,18 @@ name: bazel-tools # Name of the image to be built
...
       # we use. We don't use the Node version directly because it is not a valid
       # Docker tag.
-      DOCKER_TAG: "10.24.0"
+      NODE_DOCKER_TAG: "10.24.0"
 
 # Image names to be tagged and pushed
 images:
 - ${_REGISTRY}/${_NAME}:${_DATE_STAMP}-${_GIT_REF}-${DOCKER_TAG}
-- ${_REGISTRY}/${_NAME}:latest-${DOCKER_TAG}
+- ${_REGISTRY}/${_NAME}:bazel${BAZEL_VERSION}-go${GO_VERSION}-node${NODE_DOCKER_TAG}
```
wallrj, line 20: Maybe just say a bit about how this new tag will be used. 
I can't find any existing references to `latest-${DOCKER_TAG}` in the github.com/jetstack organization.
Will there be a follow up PR which uses this new tag?

wallrj (approved):
Still a bit confused about how this will be used.

Unhold if you need this quickly, otherwise add an explanatory comment.

/lgtm
/hold

images/bazel-tools/build.yaml
irbekrm: No, I think this is only used in a [private project](https://github.com/jetstack/fips-builds/blob/master/.github/workflows/builder-image.yaml#L24)... that currently uses sha, but I'd like to change that to use these tags instead so it's easier for me to see what versions of things are currently used.

irbekrm: I'll add a note about that there

PR #606, opened 2021-10-29
Remove old tests + run release-next periodics less frequently

SgtCoDFish:
See individual commit messages for more information

See [this comment](https://github.com/jetstack/testing/pull/605#discussion_r739222637) for more details about testing on `release-1.7`


config/jobs/cert-manager/release-next/cert-manager-release-next-periodics.yaml
```diff
@@ -1,7 +1,7 @@
 periodics:
 
 - name: ci-cert-manager-next-bazel
-  interval: 2h
+  interval: 168h  # 1 week
```
irbekrm, line 4: I'm not sure if there is a value in running a test once a week- I think usually no-one really checks a single failed test, so it's important for them to be frequent enough to be able to spot something that is not a flake by seeing a succession of failures. So I'd argue that it would be better to either run it perhaps daily or not run it at all (would prefer daily)

irbekrm (approved):
Looks good to me 👍🏼 

I've left a comment on the changed 'release-next' periodic, but it's optional, feel free to unhold if you disagree

/lgtm

config/jobs/cert-manager/release-next/cert-manager-release-next-periodics.yaml
SgtCoDFish: Yeah I take your point for sure!

I think I'd maybe prefer to just outright remove the tests for now because I think they're all _literally_ useless until we update the release-1.7 branch for the first alpha. I basically only set it to 1 week so it would be easy to change it back to 2 hours when we _do_ update the release-1.7 branch :joy: 

I'll update to remove; it'll be easy enough to re-add later!

SgtCoDFish: also... more red :imp:

irbekrm: > also... more red

where do you see the red 👀 ?

>  I basically only set it to 1 week so it would be easy to change it back to 2 hours when we do update the release-1.7 branch

I wonder if keeping the jobs in, but leaving out the interval would work, I guess that would make putting it back easier.
lgtm-ing with a hold in case you agree

/lgtm
/hold

SgtCoDFish: :grin: 

red

> I wonder if keeping the jobs in, but leaving out the interval would work, I guess that would make putting it back easier.

I'd assume that'd be a syntax error, although I've not checked! I'll take the hit on adding the tests back in if needs be, in any case! Thanks for the review! :grin:

irbekrm (approved):

PR #611, opened 2021-12-01
Adds ExperimentalSecretApplySecretTemplateControllerMinKubernetesVTODO to cert-manager feature gate env vars

JoshVanL:
Currently set to `true` in both feature groups to test which Kubernetes versions work with this feature.

The name currently has `vTODO` until we find out what is the minimum Kubernetes version this works with.

```release-note
NONE
```

PR #612, opened 2021-12-03
Changes the cert-manager presubmit config for feature gates to use AllAlpha=[true,false]

JoshVanL:
See https://github.com/jetstack/testing/pull/611#issuecomment-984790023

Change feature gate options to "all" so will always work, regardless of what features are registered on target PRs.

### DIFF
--- a/config/jobs/cert-manager/config.yaml
+++ b/config/jobs/cert-manager/config.yaml
@@ -85,13 +85,13 @@ presets:
     preset-disable-all-feature-gates: "true"
   env:
   - name: FEATURE_GATES
-    value: "ExperimentalCertificateSigningRequestControllers=false,ExperimentalGatewayAPISupport=false,ExperimentalSecretApplySecretTemplateControllerMinKubernetesVTODO=true"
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/config.yaml
```diff
@@ -85,13 +85,13 @@ presets:
     preset-disable-all-feature-gates: "true"
   env:
   - name: FEATURE_GATES
-    value: "ExperimentalCertificateSigningRequestControllers=false,ExperimentalGatewayAPISupport=false,ExperimentalSecretApplySecretTemplateControllerMinKubernetesVTODO=true"
```
munnerz, line 88: Should we also enable `AllBeta`?

munnerz: Feel like we may accidentally miss things as features graduate otherwise 😄

munnerz: Hmm actually ignore me - beta is enabled by default anyway :)

munnerz (approved):

PR #613, opened 2021-12-16
Ensure all cert-manager maintainers are OWNERS for cert-manager projects

SgtCoDFish:
Uses [the cert-manager OWNERS file](https://github.com/jetstack/cert-manager/blob/71de30931fc54eba59853097b1699ffdc4bbc974/OWNERS) as the source of truth (i.e., I copied that file to all OWNERS files under cert-manager and added the missing OWNERS file for csi-lib).

External maintainers aren't added in as reviewers because they need to be in the jetstack org; obviously this is an artifact of the tests being defined under the `jetstack` org and this will change when we move this stuff to `cert-manager/testing` (or similar)

This has the side effect of adding @jahrlin as an OWNER for all these too :tada: 

Most of this stuff should (obviously, I think) be an automated process for cert-manager stuff in the future, but I suspect that'll happen as a natural consequence of moving infra to the CNCF (which will in turn move these files to the cert-manager github org)


jahrlin:
Agreed, automating this would be a huge win

/lgtm

jakexks (approved):
/lgtm

SgtCoDFish:
> Does it mean that the individual OWNERS files in the various projects are not actually used by Prow or jetstack-bot? I guess we need to adjust the onboarding tasks smile

I think they _are_, for contributions to that repo, but I definitely don't have as good a handle on this stuff as I'd like. Ideally this stuff should flow naturally; as soon as someone's added to the cert-manager-maintainers google group they should automatically have access to things without us having to raise PRs like this (and for any given project under the cert-manager org there should probably not only be 1 OWNER, either)

I mentioned this kind of thing in the CNCF [annual review draft](https://github.com/SgtCoDFish/toc/pull/1/files#diff-ddf6d117edde871fb012342c4029aa7f70f24631f2561497b395ac30f4ce52feR69-R74) - there's surely prior art here that we could build on

PR #614, opened 2021-12-21
Enable periodic tests of the next release (cert-manager 1.7)

wallrj:
As requested in https://cert-manager.io/docs/contributing/release-process/#minor-releases

> Proceed to the post-release steps:
>     (initial alpha only) Create a PR on cert-manager/release in order to re-enable the next periodic tests configured in:
>     config/jobs/cert-manager/release-next/cert-manager-release-next-periodics.yaml
>     Why? Because we disable the “next” periodic tests right after a final release since next periodics are only useful after we do the first alpha (e.g., in PR 606).


irbekrm (approved):
/lgtm

PR #617, opened 2022-01-17
preset-disable-all-with-enable-certificate-addition-output-formats-feature-gates

JoshVanL:
Currently, Additional Output Format tests on cert-manager are being skipped on Kubernetes versions targeting v1.18 since it is a Alpha feature gate. This feature should work fine however on kube v1.18.

Adds preset option in cert-manger for disabling all feature, but enabling Certificates Additional Output Formats. Sets for presubmit and periodics for cert-manager.

/assign @irbekrm 

### DIFF
--- a/config/jobs/cert-manager/config.yaml
+++ b/config/jobs/cert-manager/config.yaml
@@ -88,6 +88,12 @@ presets:
     value: "AllAlpha=false"
 
 - labels:
+    preset-disable-all-output-formats-enable-feature-gates: "true"
+  env:
+  - name: FEATURE_GATES
+    value: "AllAlpha=false,AdditionalCertificateOutputFormats=true"
+
+- labels:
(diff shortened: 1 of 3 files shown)


config/jobs/cert-manager/config.yaml
```diff
@@ -87,6 +87,12 @@ presets:
   - name: FEATURE_GATES
     value: "AllAlpha=false"
 
+- labels:
+    preset-disable-all-output-formats-enable-feature-gates: "true"
```
irbekrm, line 91: Nit: this does read like 'disable all output formats and enable feature gates' whilst it does the opposite

irbekrm: The naming of these presets is quite painful which suggests something might not be right with this setup I guess..

PR #618, opened 2022-01-21
use go 1.17 for website job

jahrlin:
Needed for: https://github.com/cert-manager/website/pull/800

We need to use the same go version as cert-manager, which is 1.17


jakexks (approved):
/approve

PR #619, opened 2022-01-26
Don't restrict which release branches can run presubmits

maelvls:
It is useful for patching the non-current releases to be able to run tests on the Kubernetes versions that this old version supported. It does not make sense to limit which branches can run the optional tests anyways.

For example, given that `release-1.7` is the next release, we should still be able to open PRs on release-1.5 and release-1.6 since we still support them.

I also forced the 1.23 not to appear in the GitHub required checks when a PR targetting release-1.5 is opened, since cert-manager 1.5 does not support 1.23.

Signed-off-by: Maël Valais <mael@vls.dev>

### DIFF
--- a/config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
+++ b/config/jobs/cert-manager/release-previous/cert-manager-release-previous-presubmits.yaml
@@ -9,6 +9,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
     labels:
@@ -42,6 +43,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs 'bazel test --jobs=1 //...' using the 'experimental' Bazel version
@@ -78,6 +80,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Verifies the Helm chart passes linting checks
@@ -113,6 +116,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Verifies dependency related files are up to date
@@ -145,6 +149,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.17 cluster
@@ -203,6 +208,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.18 cluster
@@ -261,6 +267,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.19 cluster
@@ -319,6 +326,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.20 cluster
@@ -377,6 +385,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.21 cluster
@@ -436,6 +445,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       testgrid-create-test-group: 'false'
       description: Runs the end-to-end test suite against a Kubernetes v1.22 cluster
@@ -500,6 +510,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       description: Runs the E2E tests with 'Venafi TPP' in name
     labels:
@@ -562,6 +573,7 @@ presubmits:
     decorate: true
     branches:
     - release-1.6
+    - release-1.5
     annotations:
       description: Runs the E2E tests with 'Venafi Cloud' in name
     labels:


config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -518,16 +518,19 @@ presubmits:
...
+    optional: true
     max_concurrency: 4
     agent: kubernetes
     decorate: true
     branches:
     - master
-    - release-1.7
+    - release-.*
+    skip_branches:
+    - release-1.5
```
maelvls, line 533: Oops, `branches` and `skip_branches` are mutually exclusive: https://github.com/kubernetes/test-infra/blob/151f4dbb9106a20a901ae7b350c4fb3b1e4ee17c/prow/config/jobs_test.go#L93

config/jobs/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -518,16 +518,19 @@ presubmits:
           value: "1"
 
     # This is the default e2e test for all PRs.
-  - name: pull-cert-manager-e2e-v1-23
+  - &default-e2e
+    name: pull-cert-manager-e2e-v1-23
     context: pull-cert-manager-e2e-v1-23
-    always_run: true
-    optional: false
+    always_run: false
```
irbekrm, line 524: I think that we want to always run this test against all PRs against current master. This change will force the PR authors to manually run a command to run the test I think?

maelvls: Yes, that's my mistake, I removed `always_run: true` but should not have.

irbekrm: Perhaps you want to create a second job where `always_run: false` and branches are set to `release-*` and keep this one restricted to just master and release-1.7?

maelvls:
I can't prevent the `always_run: true` job from running on `branch-1.5`. I cloned the test-infra repo (where Prow is) and added two test cases:

```go
		{
			name: "job with always_run: true on a skipped branch should not run",
			job: Presubmit{
				Brancher: Brancher{
					SkipBranches: []string{"skipped"},
				},
				AlwaysRun: true,
			},
			ref:         "skipped",
			expectedRun: true,
		},
		{
			name: "job with always_run: true on a branch not listed in 'branches' should not run",
			job: Presubmit{
				Brancher: Brancher{
					Branches: []string{"master"},
				},
				AlwaysRun: true,
			},
			ref:         "not-listed-in-the-branches-filter",
			expectedRun: true,
		},
```

and they both failed:

```
--- FAIL: TestPresubmitShouldRun (0.00s)
    --- FAIL: TestPresubmitShouldRun/job_with_always_run:_true_on_a_skipped_branch_should_not_run (0.00s)
        jobs_test.go:1011: job with always_run: true on a skipped branch should not run: did not determine if job should run correctly, expected true but got false
    --- FAIL: TestPresubmitShouldRun/job_with_always_run:_true_on_a_branch_not_listed_in_'branches'_should_not_run (0.00s)
        jobs_test.go:1011: job with always_run: true on a branch not listed in 'branches' should not run: did not determine if job should run correctly, expected true but got false
```
which explains why the job `v1-23` runs on all the branches, not only `master` and `release-1.7`.

config/jobs/cert-manager/cert-manager-presubmits.yaml
maelvls: Unfortunately, the `always_run` flag doesn't look at the `branches`. If I had two separate jobs


```yaml
  - always_run: true
    optional: false
    branches:
    - master
    - release-1.7
  - always_run: false
    optional: true
    skip_branches:
    - release-1.5
```

then the first job would still be triggered on release-1.5 due to `always_run`. I think we are misusing `always_run`. Shouldn't we rather use `optional: false` so that the job is run for the branches listed in `branches`?

maelvls: Do I understand correctly:

- `optional: true` means that this job must report a GitHub check ([docs](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#requiring-jobs-for-auto-merge-through-tide)),
- `always_run: true` means that this job will be automatically run when a PR is opened.

I definitely think `always_run` is buggy since it doesn't account for `branches` nor `skip_branches`.

maelvls: To see if the `always_run: true` behavior was correct, I cloned the test-infra repo (where Prow is) and added two test cases:

```go
{
	name: "job with always_run: true on a skipped branch should not run",
	job: Presubmit{
		Brancher: Brancher{
			SkipBranches: []string{"skipped"},
		},
		AlwaysRun: true,
	},
	ref:         "skipped",
	expectedRun: false,
},
{
	name: "job with always_run: true on a branch not listed in 'branches' should not run",
	job: Presubmit{
		Brancher: Brancher{
			Branches: []string{"master"},
		},
		AlwaysRun: true,
	},
	ref:         "not-listed-in-the-branches-filter",
	expectedRun: false,
},
```

and the tests passed. So `always_run: true` is working as expected.

irbekrm: > Unfortunately, the always_run flag doesn't look at the branches. If I had two separate jobs

I'm not sure I understand what you mean. I meant keeping the existing job as in and adding one more that has both `always_run: false` and `optional: true` as well as `release-*`. The first one should then work in the same way as now (i.e run only against PRs for `master` and `release-1.7` and the second one should work in the same way as other modified jobs in this PR (i.e run on a PR against any branch, but only if triggered). Or, which one of those would you expect to misbehave?

maelvls: I think I was confused by the fact that I noticed that `pull-cert-manager-e2e-v1-23` was showing as a mandatory check on my PR against `release-1.5`:

Screenshot from 2022-01-26 15-18-40


I just realized that this mandatory check isn't related to `always_run: true` nor `optional: false`.

It is due to another Prow configuration, in [config.yaml](https://github.com/maelvls/testing/blob/785481625ab73f45d092f5a3bd345919f5ddd466/config/config.yaml#L52-L67):

```yaml
    jetstack:
      repos:
        cert-manager:
          protect: true
          required_status_checks:
            contexts:
            - dco
            - pull-cert-manager-bazel
            - pull-cert-manager-deps
            - pull-cert-manager-chart
            - pull-cert-manager-e2e-v1-23
```

I will do as you suggested.

maelvls: @irbekrm I ended up just adding `release-1.5` to the previous presubmits.

irbekrm (approved):
I think this makes sense for now 👍🏼 

/lgtm

PR #620, opened 2022-01-27
Bump prow configuration after cert-manager 1.7 release

wallrj:
Based on:

* https://github.com/jetstack/testing/pull/605

But in addition I've
* disabled the release-next periodics (as described in https://cert-manager.io/docs/contributing/release-process/)
* Updated the K8S versions in the release-previous presubmits to match the requirements of both 1.6 and 1.7

PR #622, opened 2022-01-28
Re-enable the release-next periodic tests

wallrj:
I commented out the release-next periodic tests in #620 but that seems to have broken the Kubernetes testgrid.
So I re-enable it but with a 1week interval.

Documented in:
 * https://github.com/cert-manager/website/pull/827

Fixes:
 * https://github.com/kubernetes/test-infra/pull/25035


irbekrm:
We did once remove them entirely (rather than commenting out) in https://github.com/jetstack/testing/pull/606
However, I guess it is possible that that change was never merged to upstream test-infra.

I guess we should have some verifier script that could have caught this error before merging.

/lgtm
/approve

irbekrm (approved):

PR #623, opened 2022-01-31
Set's `FEATURE_GATES_CA_INJECTOR` in same style and place as `FEATURE_GATES`

JoshVanL:
See https://github.com/jetstack/cert-manager/pull/4810 for context.

**Should be merged directly after but only once https://github.com/jetstack/cert-manager/pull/4810 has been merged**


JoshVanL:
/close
As per https://github.com/cert-manager/cert-manager/pull/4810#issuecomment-1027018956

PR #625, opened 2022-02-08
upgrade go version to 1.17 for webhook-example

roytev:
Signed-off-by: Roi Teveth <roytev@gmail.com>

PR #628, opened 2022-02-09
Simplify the branches list

wallrj:
Ostensibly for consistency with other cert-manager jobs, e.g.

https://github.com/jetstack/testing/blob/d903cb91142155db80ab23b3f36afdae11331aed/config/jobs/cert-manager/csi-lib/cert-manager-csi-lib-presubmits.yaml#L10-L11

But really this is a pointless and hopefully non-breaking change designed to trigger the config-updater plugin to redeploy the job-config.
It failed to do so for #625.

It may even be worse, since the branches are supposed to be a Regexp, but it's not clear from the docs whether they are a regexp match or a search:
 * https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#how-to-configure-new-jobs


irbekrm (approved):
/lgtm

Issue #629, opened 2022-02-09
Migrate cert-manager testgrid integration from Transfigure to Config Merger


wallrj:
Transfigure is deprecated in favor of [Config Merger](https://github.com/kubernetes/test-infra/blob/master/testgrid/merging.md). See [Migration](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure#migration) for details.

And Transfigure sometimes fails with false errors caused by problems in kubernetes/test-infra, preventing us merging unrelated changes to this repo, 
 * https://github.com/jetstack/testing/pull/628#issuecomment-1033756702

Istio have already completed the migration. Here's their issue: 
 * https://github.com/istio/test-infra/issues/3533

[TODO](https://github.com/kubernetes/test-infra/tree/master/testgrid/cmd/transfigure#migration)
- [x] Add  Configurator jobs (like [this example](https://github.com/kubernetes/test-infra/blob/master/testgrid/config-merger-prowjob-example.yaml)) to Prow instance.
   - [x] #673
   - [x] #676
- [x] In the same PR, in this repository, add your instance to the [mergelists](https://github.com/kubernetes/test-infra/blob/master/config/mergelists) and delete the gen-config.yaml file that Transfigure has been maintaining.
   - [x] https://github.com/kubernetes/test-infra/pull/26209
- [x] Delete any leftover Transfigure jobs and close any leftover Transfigure PRs.

inteon:
I think this has been completed, also the jetstack tests have been removed from https://testgrid.k8s.io/ (see https://github.com/kubernetes/test-infra/pull/29269).

Issue #499, opened 2021-05-26
RBAC Rules for trusted prowjobs


simplyzee:
We have different workloads that run in different clusters e.g. trusted. This YAML snippet applies the required roles to ensure the services that talk from other clusters can read, update and delete CI jobs that have run in the other clusters.

I've attached it here for now since we don't have a place to store or run apply these changes to another cluster. This YAML snippet has already been applied to -trusted and the -gke cluster 

```
kind: Role
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: test-pods
  name: "sinker"
rules:
  - apiGroups:
    - ""
    resources:
    - pods
    verbs:
    - list
    - watch
    - patch
    - get
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1
metadata:
  namespace: test-pods
  name: "sinker"
roleRef:
  apiGroup: rbac.authorization.k8s.io
  kind: Role
  name: "sinker"
subjects:
- kind: User
  name: "client"
  apiGroup: rbac.authorization.k8s.io
```

PR #630, opened 2022-02-09
WIP: Upload TestGrid config directly to GCS

wallrj:
Part of:
 * https://github.com/jetstack/testing/issues/629

Based on 
 * https://github.com/istio/test-infra/pull/3534


chases2 (approved):
After you create the bucket and service account, this all looks good to me.

prow/cluster/prow-controller-manager_rbac.yaml
```diff
@@ -111,3 +111,13 @@ subjects:
 - kind: ServiceAccount
   name: "prow-controller-manager"
   namespace: default
+---
+apiVersion: v1
+kind: ServiceAccount
```
irbekrm, line 116: I think this will create service account in Prow build cluster whereas the job actually runs in 'trusted' Prow workers cluster?
If I understand correctly, we don't have any automation at the moment to create Kubernetes resources in the trusted cluster

config/jobs/testing/testing-presubmits.yaml
```diff
@@ -78,7 +78,7 @@ presubmits:
     trigger: "(?m)^/test verify,?(\\s+|$)"
     rerun_command: "/test verify"
 
-  - name: pull-testing-check-testgrid-config
+  - name: pull-testing-check-testgrid-config-transfigure # TODO(#629) Delete this job
```
irbekrm, line 81: Suggestion: delete this job now as it cannot succeed now anyway due to recent changes to test-infra repo https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1651582675753379?thread_ts=1651581240.979269&cid=CDEQJ0Q8M

irbekrm (changes requested):
Thanks for working on this @wallrj !

I think we could now merge this as the resources have now been created in GCP.

This looks good to me as I see that someone who understands Config Merger better has approved it- my only suggestion is to remove the SA definition as it looks to me that it will be created in the wrong cluster.
I think we might need to cut the SA by hand for now and figure out how we actually want to apply Kubernetes resources to that cluster. We already have a precedent of applying resources to this cluster by hand for the same reason https://github.com/jetstack/testing/issues/499

config/jobs/testing/testing-presubmits.yaml
wallrj: Done.

config/jobs/testing/testing-trusted.yaml
```diff
@@ -60,7 +60,7 @@ presets:
 postsubmits:
   jetstack/testing:
 
-  - name: post-testing-upload-testgrid-config
+  - name: post-testing-upload-testgrid-config-transfigure # TODO(#629) Delete this job
```
wallrj, line 63: I'll remove this one too.

prow/cluster/prow-controller-manager_rbac.yaml
wallrj: Removed.

config/jobs/testing/testing-trusted.yaml
wallrj: Done.

irbekrm (approved):
Thank you for working on this!

This will be a major improvement as we will be able to keep our TestGrid config up to date now.
/lgtm

PR #639, opened 2022-03-02
Explicitly specify transfigure version

irbekrm:
Transfigure is actually deprecated and the structure of the component seems to have changed, so that our `pull-testing-check-testgrid-config` jobs are now failing (the tests failed because they pulled the latest version due to us not specifying version explicitly)

Richard is working on replacing transfigure with config-merger in https://github.com/jetstack/testing/pull/630, so this PR is just a quick fix so we can keep using the last know version of transfigure till the proper fix gets implented.

### DIFF
--- a/config/jobs/testing/testing-presubmits.yaml
+++ b/config/jobs/testing/testing-presubmits.yaml
@@ -90,7 +90,7 @@ presubmits:
       testgrid-create-test-group: "false"
     spec:
       containers:
-      - image: gcr.io/k8s-prow/transfigure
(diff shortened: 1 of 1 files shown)


irbekrm:
/override pull-testing-check-testgrid-config

😆

irbekrm:
This should fix a need to override `pull-testing-check-testgrid-config` like in https://jetstack.slack.com/archives/C81G5AR8E/p1644485788036659

SgtCoDFish (approved):
/lgtm
/approve

I can see this isn't the newest version available, but it's definitely a version that is available; looks good! :+1:

PR #641, opened 2022-03-02
Bump prow

irbekrm:
I have already upgraded Prow to the latest version (since the upgrade is anyway a manual process).
This PR:
- ensures the version change is synced to Git
- updates upgrade instructions in Readme and removes a bunch of legacy instructions

I've checked the Prow logs and testgrid https://testgrid.k8s.io/jetstack - a few tests have run since the upgrade and all seems to be working.
There are a bunch of upgrades to various parts of out CI setup which I will not do as part of this PR.


prow/version
```diff
@@ -1 +1 @@
-v20210409-985ef5e721
+v20220301-2a7bee07aa
```
irbekrm: Newline gets re-added in the next commit :)

irbekrm:
> we'll maybe have to revisit a bit given the README talks about bazel

Thanks Ashley, I think it makes sense to deploy our Prow by other means than Bazel given that we're removing Bazel elsewhere.
I guess that work won't be part of https://github.com/cert-manager/cert-manager/issues/4712 though

SgtCoDFish:
> > we'll maybe have to revisit a bit given the README talks about bazel
> 
> Thanks Ashley, I think it makes sense to deploy our Prow by other means than Bazel given that we're removing Bazel elsewhere. I guess that work won't be part of [cert-manager/cert-manager#4712](https://github.com/cert-manager/cert-manager/issues/4712) though

I guess we can cross that bridge after #4712 is done :grin:

PR #643, opened 2022-03-11
Add optional presubmit jobs "make-test" and "make-e2e-v1-23"

maelvls:
|--|

I am working on `make e2e` in https://github.com/cert-manager/cert-manager/pull/4914. I need to test that the unit tests, integration tests and end-to-end tests are passing with the new Make flow.

We will keep both Bazel and Make presubmits until the move is complete.

In this PR I added two Make presubmits to test that everything works and with the goal of releasing 1.8.0-alpha.0. Right now, these two presubmits are "optional" to avoid disruption in other PRs until https://github.com/cert-manager/cert-manager/pull/4914 is merged. That means that these two new presubmits have to be manually triggered with:

  /test pull-cert-manager-make-test
  /test pull-cert-manager-make-e2e-v1-23

Signed-off-by: Maël Valais <mael@vls.dev>

### DIFF
--- a/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
@@ -59,7 +59,7 @@ presubmits:
         # TODO: change to a custom image that embeds the system tools we need
         # (jq, make, bash, Go, etc). Tracked at
         # https://github.com/cert-manager/cert-manager/issues/4939.
-      - image: 1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
+      - image: docker.io/library/golang:1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
         args:
         - runner
         - make
@@ -642,7 +642,7 @@ presubmits:
         # TODO: change to a custom image that embeds the system tools we need
         # (jq, make, bash, Go, etc). Tracked at
         # https://github.com/cert-manager/cert-manager/issues/4939.
-      - image: 1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
+      - image: docker.io/library/golang:1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
         args:
         - runner
         - make/ci.sh 1.23
--- a/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
@@ -36,6 +36,38 @@ presubmits:
           - name: ndots
             value: "1"
 
+  - name: pull-cert-manager-make-test
+    always_run: false
+    optional: true
+    context: pull-cert-manager-make-test
+    max_concurrency: 8
+    agent: kubernetes
+    decorate: true
+    branches:
+    - master
+    - release-1.8
+    annotations:
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: jetstack-cert-manager-presubmits-blocking
+      description: Runs 'bazel test --jobs=1 //...'
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: 1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
+        args:
+        - runner
+        - make
+        - test
+        resources:
+          requests:
+            cpu: 2
+            memory: 4Gi
+      dnsConfig:
+        options:
+          - name: ndots
+            value: "1"
+
   - name: pull-cert-manager-bazel-nocache
     description: Run cert-manager unit tests with Bazel remote-caching disabled
     always_run: false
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -36,6 +36,38 @@ presubmits:
           - name: ndots
             value: "1"
 
+  - name: pull-cert-manager-make-test
+    always_run: false
```
SgtCoDFish, line 40: suggestion: Do we not actually want `always_run: true`? I think ideally we'd get to a point where every PR is tested fully with bazel and make, before we remove bazel.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -36,6 +36,38 @@ presubmits:
...
+    max_concurrency: 8
+    agent: kubernetes
+    decorate: true
+    branches:
+    - master
+    - release-1.8
+    annotations:
+      testgrid-create-test-group: 'true'
+      testgrid-dashboards: jetstack-cert-manager-presubmits-blocking
+      description: Runs 'bazel test --jobs=1 //...'
```
SgtCoDFish, line 52: suggestion: this annotation seems to be out of date

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -36,6 +36,38 @@ presubmits:
...
+    labels:
+      preset-service-account: "true"
+    spec:
+      containers:
+      - image: 1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
```
SgtCoDFish, line 57: question: Is this image missing a repo? Is it `docker.io/library/golang`?

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -36,6 +36,38 @@ presubmits:
...
+        args:
+        - runner
+        - make
+        - test
```
SgtCoDFish, line 61: suggestion: I suggest the `test-ci` target since that'll generate JUnit XML output which we can wire up to prow for test output (e.g. in [pages like this](https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/pr-logs/pull/cert-manager_cert-manager/4925/pull-cert-manager-bazel/1500889334834794496)). That might take some plumbing to get it working but I think this is the right target in any case.

```suggestion
        - runner
        - make
        - -f
        - make/Makefile
        - test-ci
```

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -578,6 +610,63 @@ presubmits:
...
+      preset-cloudflare-credentials: "true"
+      preset-retry-flakey-tests: "true"
+      preset-enable-all-feature-gates: "true"
+      preset-ginkgo-skip-default: "true"
+    spec:
+      containers:
+      - image: 1.17.8-alpine@sha256:e2e68a9cdd5da82458652fdac3908a3a270686b38039f2829855398e2e06019d
+        args:
+        - runner
+        - make/ci.sh 1.23
```
SgtCoDFish, line 641: suggestion: does this need the makefile path?
```suggestion
        - runner
        - make
        - -f
        - make/Makefile
        - ...
```

SgtCoDFish (changes requested):
Thanks for doing this!

There's also `make -f make/Makefile ci-presubmit` which will run checks, but that also adds goimports which isn't run by bazel, so I don't think we need to add that _yet_.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: You are correct. Since https://github.com/cert-manager/cert-manager/pull/4914 hasn't been merged to master yet, I propose to set the job to optional to prevent any breakage in other PRs.

I have added a TODO in 3aba2cf29694a385bb6b5b2c01fe365acf893ac8.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: Fixed in 9cf858c.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
SgtCoDFish: I agree it should be optional currently, but I think it could be `always_run: true` today at least for `make test-ci`. That code _has_ already merged to master. The _e2e_ would have to be `always_run: false` but I think we _can_ always run the tests

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: Fixed in f69b0df, thanks.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: In https://github.com/cert-manager/cert-manager/pull/4914, I have decided to replace the root Makefile with what used to be `make/Makefile`. This won't break any existing Bazel job: the root Makefile's targets weren't used anywhere.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: Thanks. This is done in a9e6708

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: @SgtCoDFish I was wrong, the Bazel Prow jobs actually use the old makefile.

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
SgtCoDFish: In the latest version I'm looking at now, the description here still refers to bazel; is there a missing commit?

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: I have re-introduced the old targets, and have added deprecation warnings in https://github.com/cert-manager/cert-manager/pull/4914/commits/fc2a4273e09bb86e18d2122840c074126802ed90. It looks a bit like this:

```make
# The targets (verify_deps, verify_chart, verify_upgrade, and cluster) are
# temorary and exist to keep the compatibility with the following Prow jobs:
#
#   pull-cert-manager-chart
#   pull-cert-manager-deps
#   pull-cert-manager-upgrade
#
# These targets should be removed as soon as the four above jobs and scripts are
# updated to use the "make" flow.
.PHONY: verify_deps
verify_deps:
	$(warning "The 'verify_deps' target is deprecated and will be removed soon. This target is not useful anymore with the new make flow.")
	./hack/verify-deps.sh

# requires docker
.PHONY: verify_chart
verify_chart:
	$(warning "The 'verify_chart' target is deprecated and will be removed soon. Please use instead 'verify-chart'.")
	bazel build //deploy/charts/cert-manager
	./hack/verify-chart-version.sh bazel-bin/deploy/charts/cert-manager/cert-manager.tgz

.PHONY: verify_upgrade
verify_upgrade:
	$(warning "The 'verify_upgrade' target is deprecated and will be removed soon. Please use instead 'make e2e-setup-kind && ./hack/verify-upgrade.sh'.")
	./hack/verify-upgrade.sh

.PHONY: cluster
cluster:
	$(warning "The 'cluster' target is deprecated and will be removed soon. Please use instead 'make e2e-setup-kind'.")
	./devel/ci-cluster.sh
```

maelvls:
/hold 

I need to sort things out. I'll let you know when done.

maelvls:
I made sure to keep the make targets that are need for the three Prow jobs (deps, chart, and upgrade).

We will be able to migrate these three to the new make flow later on.

/unhold

config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
maelvls: Oops, I missed this `description:`. 06ca5be should be the right one.

SgtCoDFish (approved):
/lgtm
/approve

I find it difficult to think how to test prow updates like this, but I think for sure this is safe to merge. If anything else needs changing I can review any future PRs for this too.

PR #645, opened 2022-03-14
Presubmits: apk add tar and grep into the busybox alpine images

maelvls:
grep and tar in BusyBox are lacking some features that we need:

- BusyBox's `grep` lacks `--null-data`,
- BusyBox's `tar` lacks `--append`.

To test that it works, I have run:

```sh
docker run -it --rm --net=host -v $PWD:/pwd -v /var/run/docker.sock:/var/run/docker.sock \
  -v $HOME/.kube:/root/.kube golang:1.17.8-alpine sh -c \
    "cd /pwd && apk add bash make curl python3 perl jq git docker tar grep && make e2e-ci K8S_VERSION=1.23"
```

and

```
docker run -it --rm --net=host -v $PWD:/pwd -v /var/run/docker.sock:/var/run/docker.sock \
  -v $HOME/.kube:/root/.kube golang:1.17.8-alpine sh -c \
    "cd /pwd && apk add bash make curl python3 perl jq git docker tar grep && make test-ci"
```

I should have tested it before merging #643 😅

cc @SgtCoDFish

### DIFF
--- a/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
@@ -56,14 +56,16 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-        # TODO: remove the "apk add" command and change to a custom image that
-        # embeds the system tools we need (jq, make, bash, Go, etc). Tracked at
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

:+1:

PR #646, opened 2022-03-15
Presubmits: for now, keep using the bazelbuild image

maelvls:
|--|

This is another follow-up to #643 and #645. I realized that Docker in docker isn't as easy as I thought, and I can't just throw the alpine image as-is.

@SgtCoDFish


SgtCoDFish (approved):
/lgtm
/approve

Untested but happy to keep reviewing :+1:

PR #647, opened 2022-03-15
Presubmits: another follow-up, remove "sudo" and add "-j"

maelvls:
|--|

This is another follow-up to #643, #645, and https://github.com/jetstack/testing/pull/646. I realized that `sudo` isn't needed, and I sped up the whole thing by adding "-j" that I had forgotten to add.

Sorry!!


SgtCoDFish (approved):
/lgtm

As before, approving without testing! :+1:

PR #649, opened 2022-03-23
Change test matrix for cert-manager-previous presubmits

SgtCoDFish:
cert-manager 1.6 doesn't support k8s 1.23


jakexks:
I just double-checked the prow documentation to make sure you could have 2 jobs with the same name, and it seems you can.

/lgtm

PR #654, opened 2022-04-01
Make sure that the "runner" command passes SIGINT down to the command that it calls (bazelbuild image)

### LINKED ISSUE #650: No kind logs when a job times out after 2 hours
When the Prow runner times out (2 hours), two unwanted behaviors occur:

1. The container logs (e.g., pebble) aren't uploaded to the "artifacts" because `make kind-logs` does not seem to be called,
2. The Prow UI doesn't show the test cases that passed and failed because Ginkgo does not produce any XML output.

Initially, I thought I had forgotten to set a trap for `SIGINT` that gets sent when the timeout is reached, because Ginkgo seemed to keep going without stopping. But in reality, these two symptoms are due to the [`runner` script](https://github.com/jetstack/testing/blob/cb6f068d60b4819fef0f9521a9c4f849d4d97928/images/bazelbuild/runner) not retransmitting SIGINT it to its children when it receives SIGINT. This is because bash ignores any signals while a child process is executing with `command "$@"`.

Here is what a Prow job looks like:

```yaml
command: /tools/entrypoint
args:
  - runner
  - bash
  - -c
  - |
    apt-get install jq -y >/dev/null
    make -j vendor-go e2e-ci K8S_VERSION=1.23
```

Let us mimick what `runner` does:

```sh
$ bash -c 'echo $$; command bash -c "trap \"echo CLEANING UP\" EXIT; sleep 1000"' &
# <----------------------------------------------------------------------------->
#                 Mimicks the command "runner".
#                  <------------------------------------------------------------>
#                  Mimicks the command "bash -c apt-get..."
#                  that is run inside the "runner" script with
#                  the line "command $@".
```

It should output the PID of the inner bash process:

```text
2896855
```

Now, type:

```sh
kill -s INT 2896855
```

Nothing happens, the bash script should still be running. The outer bash script has not passed the SIGINT signal down to its children.

But with SIGTERM, it seems to work:

```console
$ kill -s TERM 2896855
CLEANING UP
[1]  + 2896855 terminated  bash -c
```

After fixing this issue, I suggest reducing the overall timeout from 2 hours to 40 minutes.

maelvls:
Fixes #650.


images/bazelbuild/runner
```diff
@@ -88,24 +85,35 @@ if [[ "${DOCKER_IN_DOCKER_ENABLED}" == "true" ]]; then
...
+"$@" &
+
+# Bash does not "trikle down" signals. If the Bash script receives SIGINT
+# and that the above command "$@" is running, then SIGINT won't be passed
+# down to the "$@" command. To work around that, we trap SIGINT and pass it
+# down explicitely.
+#
+# shellcheck disable=SC2064
+trap "kill INT $!" INT
+wait $!
```
maelvls, line 103: This is the fix.

images/bazelbuild/runner
```diff
@@ -42,44 +41,42 @@ cleanup_binfmt_misc() {
...
     echo "Remaining docker images and volumes are:"
     docker images --all || true
     docker volume ls || true
-    # cleanup binfmt_misc
-    echo "Cleaning up binfmt_misc ..."
-    # note: we run this in a subshell so we can trace it for now
-    (set -x; cleanup_binfmt_misc || true)
+
+    # We run this in a subshell so we can trace it.
+    cleanup_binfmt_misc || true
```
inteon, line 52: is `set -x` not required anymore

images/bazelbuild/runner
inteon: why is only SIGINT passed and not SIGTERM?

images/bazelbuild/runner
maelvls: `set -x` had been added a long time ago because we had issues with vfs, but we have not had this issue since then, so I would be enclined to disable this verbose output since it doesn't seem needed anymore.

images/bazelbuild/runner
maelvls: That is a good question, I don't know why I restricted myself to SIGINT. We should also support SIGTERM since SIGTERM is probably what is being sent to processes when the Google Cloud VM is preempted.

maelvls: I have added SIGTERM in b14f5947346ce1b1934a1bc518a7c46b5ab52a38.

images/bazelbuild/runner
inteon: ok thx

inteon:
/lgtm
/hold
Lgtm, but @jahrlin might still want to review

jahrlin (approved):
/lgtm
/unhold

Thanks @maelvls !

images/bazelbuild/runner
```diff
@@ -88,24 +83,48 @@ if [[ "${DOCKER_IN_DOCKER_ENABLED}" == "true" ]]; then
...
+#  | Google Cloud VM preempted using ACPI shutdown            | SIGTERM [2] |
+#  | GKE worker removed due to scale down using ACPI shutdown | SIGTERM [2] |
+#
+#  [1]: https://github.com/kubernetes/test-infra/blob/ee1e7c8/kubetest/process/process.go#L202
+#  [2]: https://unix.stackexchange.com/questions/499761/what-signal-is-sent-to-running-programs-scripts-on-shutdown
+#
+# shellcheck disable=SC2064
+trap "kill INT $!" INT
+# shellcheck disable=SC2064
+trap "kill TERM $!" TERM
```
maelvls, line 113: I just realized that these `kill` commands are incorrect. It should be:

```sh
trap "kill -s INT $!" INT
trap "kill -s TERM $!" TERM
```

PR #655, opened 2022-04-01
make-dind is a new image for rootless dind-enabled Prow jobs

maelvls:
See https://github.com/jetstack/testing/issues/656

To test it yourself locally, you will have to configure your CRI so that it runs as non-root:

- **Nerdctl:** read https://github.com/containerd/nerdctl/blob/main/docs/rootless.md
- **Docker:** https://rootlesscontaine.rs/getting-started/docker

Then, build the dind image (choose the correct command):

```bash
nerdctl build -t local/make-dind -f images/make-dind/Dockerfile ./images/make-dind
docker build -t local/make-dind -f images/make-dind/Dockerfile ./images/make-dind
```

Finally, run the image:

```
nerdctl run -it --privileged -v /tmp:/tmp --tmpfs /var/lib/containerd local/make-dind bash
docker run -it --privileged -v /tmp:/tmp --tmpfs /var/lib/containerd local/make-dind bash
```

You should get a prompt. Try to run a container:

```bash
nerdctl run alpine ls
```

> ⚠️ The following warning can be ignored:
> ```
> WARN[0002] default network named "bridge" does not have an internal nerdctl ID or nerdctl-managed config file, it was most likely NOT created by nerdctl 
> ```

Why `--privileged` when we want a rootless container? The container requires a number of capabilities to configure cgroups for the child containers and such.

### DIFF
--- a/images/make-dind/Dockerfile
+++ b/images/make-dind/Dockerfile
@@ -1,0 +1,15 @@
+FROM debian
+
+# Usage:
+#
+#   nerdctl build -t local/make-dind -f images/make-dind/Dockerfile ./images/make-dind
+#   nerdctl run -it --privileged -v /tmp:/tmp --tmpfs /var/lib/containerd local/make-dind bash
+
+RUN apt update \
+    && apt install curl bash make curl python3 perl jq git tar grep iptables -y \
+    && curl -sSL https://github.com/containerd/nerdctl/releases/download/v0.17.1/nerdctl-full-0.17.1-linux-amd64.tar.gz | tar zx -C / \
+    && mkdir -p /opt/cni/bin \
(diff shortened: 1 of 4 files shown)


inteon:
This is how I tested the PR (based on https://docs.docker.com/engine/security/userns-remap/):
1. update `/etc/docker/daemon.json` and add `"userns-remap": "default"`
2. reboot my system (just restarting containerd didn't seem to do the trick)

maelvls:
Everything in this PR is pointless: we don't have control over the CRI!! Since we are using GKE and that the GKE CRI is root, there is no point in trying to make our container image "rootless-ready" when the CRI will run as root anyways...

PR #658, opened 2022-04-05
Disable SSA feature for Kubernetes 1.19 - 1.21

irbekrm:
Disables server-side apply feature for kube 1.19 - 1.21

See https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1649145566523839 for context

In long term, we should investigate the error, however it would probably take a while and in short term it is more valuable to see the e2e tests pass.


jakexks (approved):
/lgtm

PR #661, opened 2022-04-20
Fix upgrade tests

irbekrm:
This upgrade periodic test has been failing for a while now https://testgrid.k8s.io/jetstack-cert-manager-master#ci-cert-manager-upgrade

It appears that the new make target to create a cluster runs a check for some tools that may not be needed by the actual tests https://github.com/cert-manager/cert-manager/blob/master/make/tools.mk#L407-L418

I've fixed this by temporarily reverting to the old target as that way we can avoid installing unnecesary deps on every test run.
Long term solution should be to create an image with all the necessary tools embedded [cert-manager#4393](https://github.com/cert-manager/cert-manager/issues/4939), I think this work should be prioritized as by installing tools every time we consume more resources which might be contributing to the recent test flakes.


jakexks (approved):
/lgtm

PR #671, opened 2022-05-04
Removes retest ProwJob

irbekrm:
As discussed in cert-manager-dev meeting on 4th May 2022, this PR removes the retester ProwJob that was periodically re-running tests on approved PRs where the tests have failed.

It seems like in practice we didn't have any value in those tests as in almost all cases a manual action was anyway required where tests had failed after a PR has been approved and the retest bot was just creating noise and increasing our test-infra cost.

Arguably, we also shouldn't approve & lgtm PRs where the tests have not passed and if they _do_ fail a human should check why they failed.


wallrj (approved):
Thanks @irbekrm 

/lgtm

PR #633, opened 2022-02-16
Removes the legacy update_index job

irbekrm:
As I understand, [we no longer need to update Algolia index this way](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1643030591411200?thread_ts=1643018915.399200&cid=CDEQJ0Q8M) and this job is continuosuly failing https://testgrid.k8s.io/jetstack-cert-manager-website#cert-manager-website-update-index

https://github.com/cert-manager/infrastructure#algolia

Signed-off-by: irbekrm <irbekrm@gmail.com>

PR #672, opened 2022-05-04
Pin configurator to a particular version

irbekrm:
This PR attempts to pin configurator tool used to validate and upload TestGrid config to the same version that seems to be working for Istio and reference the configurator tool directly. I looked at Istio's ProwJobs [here](https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/test-infra/istio.test-infra.trusted.master.yaml#L4-L31) and [here](https://github.com/istio/test-infra/blob/master/prow/cluster/jobs/istio/test-infra/istio.test-infra.master.yaml#L3-L26)

The configurator job failed https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/pr-logs/pull/jetstack_testing/671/pull-testing-check-testgrid-config/1521898589293383680 I think that might have been because we pulled in the latest version and I guess maybe the filesystem of the container has changed.

### DIFF
--- a/config/testgrid/dashboards.yaml
+++ b/config/testgrid/dashboards.yaml
@@ -4,16 +4,12 @@ dashboard_groups:
   dashboard_names:
   - jetstack-cert-manager-master
   - jetstack-cert-manager-previous
-  - jetstack-cert-manager-next
   - jetstack-cert-manager-presubmits-blocking
-  - jetstack-cert-manager-website
   - jetstack-testing-janitors
 
 # Dashboards
 dashboards:
 - name: jetstack-cert-manager-master
 - name: jetstack-cert-manager-previous
-- name: jetstack-cert-manager-next
 - name: jetstack-cert-manager-presubmits-blocking
-- name: jetstack-cert-manager-website
 - name: jetstack-testing-janitors
--- a/config/jobs/testing/testing-presubmits.yaml
+++ b/config/jobs/testing/testing-presubmits.yaml
@@ -91,9 +91,9 @@ presubmits:
       testgrid-create-test-group: "false"
     spec:
       containers:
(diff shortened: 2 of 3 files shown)


irbekrm:
I have manually modified Prow job configs in the cluster and this is the latest error that we get with the changes in this PR https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/pr-logs/pull/jetstack_testing/672/pull-testing-check-testgrid-config/1521927458712981504

config/testgrid/dashboards.yaml
```diff
@@ -4,16 +4,12 @@ dashboard_groups:
   dashboard_names:
   - jetstack-cert-manager-master
   - jetstack-cert-manager-previous
-  - jetstack-cert-manager-next
   - jetstack-cert-manager-presubmits-blocking
-  - jetstack-cert-manager-website
```
irbekrm, line 9: We no longer should have any ProwJobs for cert-manager website that would be marked to be displayed by TestGrid since we removed the algolia check https://github.com/jetstack/testing/pull/633

config/testgrid/dashboards.yaml
```diff
@@ -4,16 +4,12 @@ dashboard_groups:
   dashboard_names:
   - jetstack-cert-manager-master
   - jetstack-cert-manager-previous
-  - jetstack-cert-manager-next
```
irbekrm, line 7: The next release jobs were (temporarily) removed altogether

wallrj (approved):
Thanks @irbekrm 

And for all the explanatory comments and links.

I cross checked this config with the Istio config you linked and agree that they now look the same, so :crossed_fingers:  that this will now work. 

/lgtm

PR #675, opened 2022-05-06
Adds a new presubmit for Kubernetes 1.24

irbekrm:
Adds an optional presubmit for Kubernetes 1.24 (copy-pasted the same for 1.23 and changed some values.

1.24 kind cluster option was already added to cert-manager in https://github.com/cert-manager/cert-manager/pull/5096 , but I _think_ we may need to PR kind cli version bump to cert-manager to have this actually working. If so, it would be good to merge this PR first as then we will be able to run this presubmit against the cert-manager PR that bumps the kind version.

This should allow us to test that the Config Merger stuff works.

### DIFF
--- a/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
+++ b/config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
@@ -649,6 +649,85 @@ presubmits:
         - name: ndots
           value: "1"
 
+  - name: pull-cert-manager-make-e2e-v1-24
+    context: pull-cert-manager-make-e2e-v1-24
+    always_run: false
+    optional: true
+    max_concurrency: 4
+    agent: kubernetes
+    decorate: true
+    branches:
+    - master
+    annotations:
+      testgrid-create-test-group: 'true'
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/cert-manager/cert-manager-presubmits.yaml
```diff
@@ -649,6 +649,85 @@ presubmits:
...
+    labels:
+      preset-service-account: "true"
+      preset-dind-enabled: "true"
+      preset-cloudflare-credentials: "true"
+      preset-retry-flakey-tests: "true"
+      preset-enable-all-feature-gates: "true"
+      preset-ginkgo-skip-default: "true"
+    spec:
+      containers:
+        # TODO: change to a custom image that embeds the system tools we
```
irbekrm, line 674: This should really be done soon, I _think_ we may be able to use [the bazel tools image](https://github.com/jetstack/testing/tree/master/images/bazel-tools) as that already has `jq` and `make`, but not sure whether we need any extra functionality to set up docker/ensure logs are exported etc

wallrj (approved):
Looks fine to me.

Let's see if the test-grid config merger runs successfully this time...

/lgtm

PR #677, opened 2022-05-09
Adds a semversort tool

irbekrm:
Updates bazel-tools image to add a script that can semantically sort tags.

I would have preferred a (tested) CLI tool, but haven't really found one.

Perhaps we could also use this in cert-manager, see i.e 

This is specifically needed in an internal project, however perhaps we coudl use this elsewhere too to be consistent i.e https://github.com/cert-manager/cert-manager/pull/5104/commits/5f4877443c1612a3bdd7f61d5e63afa611e59246#diff-78e3cf9e2d802b7bfed7e7a17262b4db83e9ff5859f84f1ace9a75169dab33eeR34-R35

Also bumps `gh` version.

### DIFF
--- a/images/bazel-tools/Dockerfile
+++ b/images/bazel-tools/Dockerfile
@@ -19,17 +19,20 @@ FROM ${BASE_IMAGE}
 
 LABEL maintainer="cert-manager-maintainers@googlegroups.com"
 
+COPY semversort.sh /usr/local/bin/semversort
+
 ARG NODE_VERSION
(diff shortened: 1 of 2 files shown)


SgtCoDFish (approved):
/lgtm
/approve

(unsurprisingly I've not really checked the various sed commands but I can review any further updates too!)

irbekrm:
> (unsurprisingly I've not really checked the various sed commands but I can review any further updates too!)

I expect this to _just work_ and to not have to _think_ about those seds :)

PR #680, opened 2022-05-12
Install jq in bazelbuild image

SgtCoDFish:
This is required by the new cert-manager make-based build system. Longer term we won't be using the "bazelbuild" image for the make build system, but having support for make-inside-bazel and make-or-bazel using this image will be helpful during the transition.


irbekrm:
Thanks for doing this!

/lgtm
/approve

irbekrm:
https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/logs/post-testing-push-bazelbuild/1524704758286258176

PR #682, opened 2022-05-12
Bump bazelbuild to new image, update Make jobs

SgtCoDFish:
We previously had a hacky way of installing jq pending its inclusion in the base image we use for test jobs. Now that we've merged https://github.com/jetstack/testing/pull/680 we don't need to do this step any more and can instead just run "make" as we'd originally intended.

This also updates bazelbuild everywhere else for cert-manager.

**ALSO** the second commit removes 1.19 testing jobs on master since we agreed not to support 1.19 in cert-manager 1.9 (as per the [supported releases page](https://cert-manager.io/docs/installation/supported-releases/) which was updated in https://github.com/cert-manager/website/pull/980 )


SgtCoDFish:
Fixes https://github.com/cert-manager/cert-manager/issues/4939

config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml
```diff
@@ -49,20 +49,17 @@ periodics:
...
-      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20210831-4cf7b0b-4.2.1
+      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220512-b6ea825-4.2.1
         args:
           - runner
-          - bash
-          - -c
-          - |
-            apt-get install jq -y >/dev/null
-            make -j vendor-go ci-presubmit test-ci
+          - make
```
irbekrm, line 58: Nice, this looks a lot cleaner 👍🏼

config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml
```diff
@@ -113,17 +110,14 @@ periodics:
...
-          - -c
-          - |
-            apt-get install jq -y >/dev/null
-            make -j vendor-go e2e-ci K8S_VERSION=1.19
+        - runner
+        - make
+        - -j
+        - vendor-go
+        - e2e-ci
+        - K8S_VERSION=1.19
```
irbekrm: I think it might make sense to pass kube version as an env var for consistency as that's how other configurable options (i.e `GINKGO_FOCUS`/`GINKGO_SKIP`) are passed
But that might be a job for another PR as I think the make scripts currently don't read them from env (?)

irbekrm (approved):
Looks good to me!

Thanks for keeping improving this bit

/lgtm

config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml
SgtCoDFish: Yeah that'd make sense, definitely something to keep in mind!

PR #685, opened 2022-05-13
Cleanup

irbekrm:
A couple cleanup fixes for 'previous' jobs:

- Use the same form of make command everywhere
- Fix `always-run` value for a job that shouldn't always run
- Remove 'context' values (as I understand from [this Prow code comment](https://github.com/kubernetes/test-infra/blob/master/prow/config/jobs.go#L285-L287) this value determines the job name in GitHub and defaults to the same as the job name). We should be always fine with the default and having it explicitly specified has no value as it is not obvious what this field does from its name and is just one more value to forget to change when copy pasting these configs around (which I did in a previous PR)


SgtCoDFish (approved):
/lgtm
/approve

Thanks :grin:

Issue #684, opened 2022-05-12
Investigate whether we can make ProwJobs configs more DRY


irbekrm:
See https://github.com/jetstack/testing/pull/683#issuecomment-1125063744 for context

Perhaps we could use a simple yaml templater like ytt and generate the job configs in the postsubmit that updates Prow config in cluster.

If we do this, we should make sure that the solution is easy to use and doesn't require people to learn a complicated tool or language and that a project that doesn't require templating can still add a plain yaml job config.

Perhaps we could use ytt or a similar templating tool.

maelvls:
Part of this issue was solve by Ashley's `cmrel generate-tests` (https://github.com/cert-manager/release/pull/80). 

The Prow configs aren't automatically PR'ed to jetstack/testing though, so I guess we still need to figure that out (cf. https://github.com/cert-manager/release/pull/94 where the automation was disabled due to permissions).

inteon:
We moved the prowgen tool to this repo and it is run as part of CI.
Additionally, with Makefile modules, most other jobs now look very similar.

PR #688, opened 2022-05-17
Adds a missing dash

irbekrm:
Small fix for a couple job configs.

In long term I think it's be good if all these didn't have to be copy pasted around as at least I make a lot of mistakes doing that- see #684 for context


jakexks (approved):
/lgtm

PR #692, opened 2022-06-07
Make for all e2e

irbekrm:
This PR sets all e2e tests (periodic and presubmit) against current master to run with make only.

At the moment some of them run with make, some with Bazel which makes it harder to investigate flakes etc.

I've also removed a couple 'experimental' Bazel jobs that we no longer need.


jahrlin (approved):
👏

PR #694, opened 2022-06-08
Revert previous jobs to their actual config on release day

munnerz:
This PR reverts changes that were made after the `v1.8.0` release to the test config that switched us to use `make` instead of the same config that was used during the development cycle.

This is the first step towards us getting periodic jobs against `release-previous` to go green again, which are currently entirely red (aside from v1.23 and v1.24): http://testgrid.k8s.io/jetstack-cert-manager-previous

As part of this PR, I've _also_ removed the v1.24 periodic job that was added post-v1.8.0 release. I'll get this added back once things are stabilised given that the changes needed to make it pass in v1.8 were already merged into the core repo's release-1.8 branch.

More context here: https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1654691889270499

cc @irbekrm

### DIFF
--- a/config/jobs/cert-manager/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
@@ -19,7 +19,7 @@ periodics:
     description: Runs 'bazel test --jobs=1 //...'
   spec:
     containers:
-    - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220512-b6ea825-4.2.1
+    - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20210831-4cf7b0b-4.2.1
       args:
       - runner
       - bazel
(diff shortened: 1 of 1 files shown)


munnerz:
> Looks good to me, but I think we should keep the 1.24 test- I think it makes sense to run the tests that were already run with Bazel with it, but the 1.24 is unrelated to that as it was added later.

👍 to keeping them there, but for now I'm really trying to just get us to a green state and then will get that added back :)

irbekrm:
The 1.24 test is mostly green https://prow.build-infra.jetstack.net/?job=ci-cert-manager-previous-e2e-v1-24

The errors that there are are the ones we still have in master currently as far as I can see.

config/jobs/cert-manager/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
```diff
@@ -174,28 +173,40 @@ periodics:
     preset-bazel-remote-cache-enabled: "true"
     preset-bazel-scratch-dir: "true"
     preset-cloudflare-credentials: "true"
-    preset-enable-all-feature-gates: "true"
+    preset-enable-all-feature-gates-disable-ssa: "true"
```
munnerz, line 176: It looks like for some reason the SSA tests were *enabled* against 1.8 even on older Kubernetes clusters... not sure why, but this could also be related to the issues we saw

munnerz:
Here's a job kicked off using `mkpj` that has passed (first time, and for the first time in a while for this particular configuration) here: https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/logs/ci-cert-manager-previous-e2e-v1-21/1534558120527794176

It tests v1.21 and release-1.8.

I have added back in the v1.24 periodic jobs in an additional commit, so I think this should be good to merge 😄

config/jobs/cert-manager/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
munnerz: Yeah some kind of configuration generator would be great - Kubernetes actually already does this today too: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes/generated/generated.yaml

Which is generated using: https://github.com/kubernetes/test-infra/tree/master/releng

config/jobs/cert-manager/cert-manager/release-previous/cert-manager-release-previous-periodics.yaml
```diff
@@ -386,6 +386,53 @@ periodics:
...
+  annotations:
+    testgrid-create-test-group: 'true'
+    testgrid-dashboards: jetstack-cert-manager-previous
+    testgrid-alert-email: cert-manager-dev-alerts@googlegroups.com
+    description: Runs the end-to-end test suite against a Kubernetes v1.24 cluster
+  labels:
+    preset-service-account: "true"
+    preset-dind-enabled: "true"
+    preset-bazel-remote-cache-enabled: "true"
+    preset-bazel-scratch-dir: "true"
```
irbekrm, line 408: These two aren't needed for make tests btw, but also shouldn't cause issues

PR #699, opened 2022-06-20
Mount download cache as `_bin`

SgtCoDFish:
This should be merged ASAP after https://github.com/cert-manager/cert-manager/pull/5130 so that we don't lose cacheing in CI when that PR merges!


jahrlin (commented):
/lgtm
/hold

SgtCoDFish:
/unhold

(Unholding as [the rename PR has merged](https://github.com/cert-manager/cert-manager/pull/5130))

PR #702, opened 2022-06-27
Don't run unit/integration tests with bazel where we use make

SgtCoDFish:
I think we could've made this change a while ago and saved some build time :sweat_smile: 

We run unit and integration tests with make for release-1.8 and master. We don't need to also run them with bazel.

We'll be able to remove the bazel tests entirely when all of our tests are available in make, which shouldn't take long. Still, we can take this win now!


jakexks (approved):
/lgtm

PR #707, opened 2022-06-29
Bump golang dind

irbekrm:
This PR:
- bumps Go versions for the various tags of `golang-dind` image we build
- bumps base image to one with a newer version of Debian and Docker
- removes some older Go versions as we shouldn't need those anymore

We should eventually automate this process so that when a newer base or other dependencies become available, there are automated PRs to update all the various images that use those dependencies. But perhaps we should first remove Bazel from the base image and see what the flow looks like then


wallrj (approved):
Thanks for sorting all this. I had no idea how all the prow images were built.

/lgtm

PR #710, opened 2022-06-30
Don't test using bazel on master

SgtCoDFish:
This leaves only the upgrade test using bazel as of now

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -36,11 +36,10 @@ branch-protection:
         cert-manager:
           required_status_checks:
             contexts:
-            - pull-cert-manager-bazel
             - pull-cert-manager-deps
             - pull-cert-manager-chart
             - pull-cert-manager-e2e-v1-24
-            # - pull-cert-manager-make-test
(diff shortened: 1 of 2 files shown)


irbekrm:
@SgtCoDFish also https://github.com/jetstack/testing/blob/master/config/config.yaml#L39-L40

SgtCoDFish:
> @SgtCoDFish also https://github.com/jetstack/testing/blob/master/config/config.yaml#L39-L40

Thank you, I always forget stuff like this!

SgtCoDFish:
@irbekrm I also added the make test as required; does it look correct now?

irbekrm:
Thanks!

/lgtm
/approve

irbekrm:
Ahh wait, do we not have to remove `pull-cert-manager-deps`?

irbekrm:
> Ahh wait, do we not have to remove pull-cert-manager-deps

I'll do that in a separate PR. Apart from that I think it was right

PR #712, opened 2022-06-30
Update upgrade test to reference new target

SgtCoDFish:
This should remove the last traces of bazel in our presubmits!

### DIFF
--- a/config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml
+++ b/config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml
@@ -318,16 +318,17 @@ periodics:
   labels:
     preset-service-account: "true"
     preset-dind-enabled: "true"
-    preset-make-volumes: "true"
     preset-default-e2e-volumes: "true"
+    preset-make-volumes: "true"
   spec:
     containers:
       - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220512-b6ea825-4.2.1
         args:
           - runner
           - make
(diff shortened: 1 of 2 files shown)


irbekrm:
Thank you very much for working on this @SgtCoDFish !

Should we make the same change [in the periodic](https://github.com/jetstack/testing/blob/master/config/jobs/cert-manager/cert-manager/cert-manager-periodics.yaml#L329-L330)? Apart from that happy to lgtm

SgtCoDFish:
Good suggestion, thanks @irbekrm  - not sure how I missed that.

I've added a second commit which does that and also which reduces the resource requirements which seemed high. Does that look reasonable? :grin:

irbekrm:
Thanks @SgtCoDFish !

For the resource requirements, I would slightly prefer if we re-evaluated what proportion of node resources each test type needs + whether we could, for example benefit from a different node size (as there probably have been some changes in that now that we use make) + monitor for possible flakes in one single PR so that all the reasoning etc is in one place.
At the moment the resource requiremement for each job that spins up kind cluster is ~50% of a node so that no more than 2 such jobs get allocated to that node.

Adding a hold if you agree else happy for it to go in as is

/hold
/lgtm
/approve

SgtCoDFish:
@irbekrm I've reverted the resource changes - I agree that it's probably better to attack them all in one go, yeah.

irbekrm:
Thanks @SgtCoDFish and thanks for working on this!

/hold cancel
/lgtm

PR #715, opened 2022-07-05
Test using cron periodics rather than intervals

SgtCoDFish:
If we move to [generating our tests](https://github.com/cert-manager/release/pull/80), we can have our generator automatically spread periodics throughout the hour using cron-style definitions intead of intervals.

This should ensure that we avoid situations where we have a tonne of tests which all run on the hour causing a spike in resources but leaving the cluster less utilized the rest of the time.

This is a manual test to check that cron-style jobs work on our deployed version of prow today.

The actual test definitions changed here were chosen arbitrarily.


jakexks (approved):
/lgtm

After a discussion on slack, we will keep an eye on this to make sure it's working and revert it if not.

PR #718, opened 2022-07-06
Mount bindir appropriately for older release tests

SgtCoDFish:
We changed the mount point to _bin for tests against master and new releases because that's the new name of the bindir

The bindir changes weren't backported into release-1.8 and so this means we don't make use of the bin/downloaded cache on release-1.8

This change will allow us to use that same cache for more dependencies


SgtCoDFish:
This is moot until we've finished testing #720 which disables caching!

SgtCoDFish:
We're gonna keep caching disabled for now, so this is entirely redundant

PR #719, opened 2022-07-06
Run upgrade test every hour

SgtCoDFish:
This is temporary to ensure the test is reliable before we release 1.9

As discussed [on slack](https://kubernetes.slack.com/archives/CDEQJ0Q8M/p1657120059312519)


SgtCoDFish:
no point

PR #722, opened 2022-07-20
Removes cert-manager/istio-csr e2e tests targeting istio version v1.7 to v1.9

JoshVanL:
The istio version v1.7 and v1.9 fail to install on Kubernetes v1.24 which is what we e2e test in.

See [here](https://github.com/cert-manager/istio-csr/pull/163).

Rather than trying to play a dance with running different kube versions with different istio versions, let's just drop support for v1.9 and below istio-csr. v1.9 has been EOL since 2022-10 anyway.


jakexks (approved):
/lgtm

PR #723, opened 2022-07-22
Use generated tests!

SgtCoDFish:
Note that there are a few handwritten tests remaining for 1.8 only since we didn't write generation code for bazel tests.

This was generated using `cmrel` from [this branch](https://github.com/cert-manager/release/pull/82). That branch should be merged first!

Then `cmrel` was fixed in [this branch](https://github.com/cert-manager/release/pull/83) which should also be merged first!


jakexks (approved):
/lgtm
I will close my eyes and trust the automation

PR #724, opened 2022-07-25
Generate tests based on branches, not modes

SgtCoDFish:
There are two parts here:

1. Updating generated tests to use branches rather than "modes". That means tests get branches embedded in their names. The same happens with dashboards.
2. Changing the dashboard names to reflect the new ones


SgtCoDFish:
Added a second commit which removes the upgrade tests for release-1.8 since they're manually specified. See [this comment](https://github.com/cert-manager/release/pull/84#issuecomment-1196643059) and the newly added comments in the commit on the cmrel tool.

config/jobs/cert-manager/cert-manager/release-1.8/cert-manager-1.8.yaml
```diff
@@ -324,48 +317,9 @@ presubmits:
...
-            - SYS_ADMIN
-      dnsConfig:
-        options:
-        - name: ndots
-          value: "1"
-    branches:
-    - release-1.8
-    - release-1.9
-    always_run: true
-    optional: false
```
SgtCoDFish, line 367: note: this test shouldn't run against release-1.8, so its removal is actually a bugfix of an issue which currently blocks backporting for release-1.8

jahrlin (approved):
Works for me and makes sense!

/lgtm

PR #728, opened 2022-08-08
Regenerate master prow jobs

inteon:


irbekrm (approved):
Thanks @inteon !

It looks to me that all the work for the Trivy tests to pass has been merged, so cannot see why to not add the new test 
https://github.com/cert-manager/release/pull/85

/lgtm

PR #729, opened 2022-08-08
Regenerate master prow files (fix wrong tag)

inteon:
Signed-off-by: Tim Ramlot <42113979+inteon@users.noreply.github.com>


irbekrm (approved):
Thanks @inteon !

I see that `docker pull eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220629-ee75d11-4.2.1` succeeds, so this should be right 👍🏼 

/lgtm

PR #743, opened 2022-08-18
update ingress class annotation for prow ingress

BeckyPauley:
This PR updates the ingress class annotation/ingress resource to match what is deployed in the cluster (changes tested and manually applied).


irbekrm (approved):
Thanks for these updates @BeckyPauley !

I've ran `bazel run //prow/cluster:production.apply` from this branch and verified that no changes get applied and https://prow.build-infra.jetstack.net/ can still be accessed

/lgtm

PR #748, opened 2022-08-30
add preStop hook to upload kind logs

jahrlin:
This PR adds a preStop-hook that will upload kind logs, to allow for these logs to be uploaded when a build times out: https://github.com/jetstack/testing/issues/650

Example:
- https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/logs/ci-cert-manager-master-e2e-v1-20/1564562018881507328

Trying this out first before I open a PR to `release` with the relevant changes

/assign @maelvls

### DIFF
--- a/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
+++ b/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
@@ -18,7 +18,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220812-2f64076-4.2.1
+      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20220830-c65cd19-4.2.1
         args:
         - runner
         - make
@@ -30,6 +30,13 @@ presubmits:
           requests:
             cpu: 2000m
             memory: 4Gi
+        lifecycle:
+          preStop:
+            exec:
+              command:
+              - /bin/sh
+              - -c
+              - make kind-logs
       dnsConfig:
         options:
         - name: ndots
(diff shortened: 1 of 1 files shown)


maelvls:
I propose to have this one merged for now. I will check whether the `postStop` can be removed or whether it has to stay, because I noticed that [`make/e2e-ci.sh`](https://github.com/cert-manager/cert-manager/blob/master/make/e2e-ci.sh) is a Bash script, which means it has the exact same problem as what [I fixed in the `runner` script](https://github.com/jetstack/testing/pull/746): the signals aren't forwarded to the make command in line 19, meaning that SIGINT and SIGTERM won't be sent to Ginkgo. But we will get the kind logs.

maelvls:
Make correctly forwards SIGTERM:

```sh
# In a shell session:
$ make -f - <<'EOF'
foo:
        sleep 1000 & trap "echo trap called; kill -s TERM $$!" TERM; wait $$!
EOF

# In another shell session:
kill -s TERM $(ps -x | grep 'make -f -' | grep -v grep | awk '{print $1}')

# ✅ The output "trap called" is shown, and make exits with 143.
```

Make does not forward SIGINT:

```sh
# In a shell session:
$ make -f - <<'EOF'
foo:
        sleep 1000 & trap "echo trap called; kill -s INT $$!" INT; wait $$!
EOF

# In another shell session:
kill -s INT $(ps -x | grep 'make -f -' | grep -v grep | awk '{print $1}')

# ❌ Nothing happens.
```

This behavior is explained in the Stack Overflow question [Why GNU make ignores SIGINT, but then doesn't execute following commands?](https://stackoverflow.com/questions/55798030/why-gnu-make-ignores-sigint-but-then-doesnt-execute-following-commands).

inteon:
@jahrlin instead of manually changing these files, the https://github.com/cert-manager/release should have been updated to generate this configuration. I'll create a PR to do this.

maelvls:
Tim showed me that the `preStop` should be on `make e2e-ci` jobs, not the `make test-ci` jobs. Currently, the `preStop` blocks that were added in this PR don't do anything.

PR #749, opened 2022-09-07
Adds nix-dind image. Docker image for nix which contains docker in docker functionality.

JoshVanL:
I would like to experiment with using nix in the `csi-lib` project for setting up the e2e dependencies. See [here](https://github.com/cert-manager/csi-lib/pull/31#issuecomment-1238257743).

### DIFF
--- a/config/jobs/testing/testing-trusted.yaml
+++ b/config/jobs/testing/testing-trusted.yaml
@@ -462,3 +462,40 @@ postsubmits:
           privileged: true
           capabilities:
             add: ["SYS_ADMIN"]
+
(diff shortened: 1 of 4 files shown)


JoshVanL:
@irbekrm I added the push testing job for nix-dind. PTAL!

irbekrm (approved):
Thanks @JoshVanL, I've not tried building this, but from a read through the code looks like it should work, this PR should update Prow's config, but am not entirely sure if it will then trigger the job as the change will have been merged already- perhaps a second PR will be needed

/lgtm

PR #761, opened 2022-09-22
Set default kubernetes version to k8s 1.24 instead of 1.25

inteon:
@jahrlin informed me that we switched to k8s 1.25 too early.
After the 1.10 cert-manager release, we will switch to kubernetes 1.25 for all our default tests.
This PR contains the changes generated by this commit: https://github.com/cert-manager/release/commit/9906389648939e3aef97c0edabd2f19c50a96620.


jahrlin (commented):
/lgtm
/approve

PR #765, opened 2022-10-07
make k8s 1.25 default

jahrlin:
This makes K8s v1.25 the default version for presubmits on master

generated from https://github.com/cert-manager/release/pull/104

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -57,8 +57,8 @@ branch-protection:
                 contexts:
                 - pull-cert-manager-master-chart
                 - pull-cert-manager-master-make-test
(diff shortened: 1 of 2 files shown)


irbekrm:
Thanks @jahrlin !

I think we also need to update which tests are required for PRs here https://github.com/jetstack/testing/blob/master/config/config.yaml#L60-L61

jahrlin:
> Thanks @jahrlin !
> 
> I think we also need to update which tests are required for PRs here https://github.com/jetstack/testing/blob/master/config/config.yaml#L60-L61

Fixed 👍

irbekrm (approved):
Thanks @jahrlin !

/lgtm

PR #770, opened 2022-10-17
use go 1.19 for website

jahrlin:

### DIFF
--- a/config/jobs/cert-manager/website/cert-manager-website-presubmits.yaml
+++ b/config/jobs/cert-manager/website/cert-manager-website-presubmits.yaml
@@ -15,7 +15,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:latest-1.19"
+      - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:20221017-bcd2286-1.19"
         args:
         - ./scripts/verify-release
         resources:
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/website/cert-manager-website-presubmits.yaml
```diff
@@ -15,7 +15,7 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:20220302-b57c609-1.17"
+      - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:latest-1.19"
```
SgtCoDFish, line 18: suggestion: I've seen some squeamishness about using any kind of "latest" within Jetstack, and I think I tend to agree with it. Do you think we should use the tag instead?

(From `crane ls eu.gcr.io/jetstack-build-infra-images/golang-nodejs`)

```suggestion
      - image: "eu.gcr.io/jetstack-build-infra-images/golang-nodejs:20221017-bcd2286-1.19"
```

SgtCoDFish (approved):
/lgtm
/hold
/approve

Happy to merge so you're not blocked but got a suggestion on the tag!

config/jobs/cert-manager/website/cert-manager-website-presubmits.yaml
jahrlin: Ah, thanks! Did not see that tag in GCR. Updated!

PR #776, opened 2022-11-23
Add trust-manager periodic for building debian bundle

SgtCoDFish:
Must be merged after https://github.com/cert-manager/trust-manager/pull/71


SgtCoDFish:
I think we'll go in a different direction to this that won't involve prow, so removing this!

PR #778, opened 2022-12-16
Apply patch from https://github.com/cert-manager/release/pull/113

SgtCoDFish:


irbekrm (approved):
Arguably we could have added the _periodic_ jobs for 1.26 in a separate PR after the PRs to add 1.26 to cert-manager master and release-1.10 are merged as else these periodics will get triggered now and go red. However, I don't see a major issue with that because we do expect them to go red, so it's just the appearance of the TestGrid :)

/lgtm

SgtCoDFish:
> Arguably we could have added the periodic jobs for 1.26 in a separate PR after the PRs to add 1.26 to cert-manager master and release-1.10 are merged as else these periodics will get triggered now and go red. However, I don't see a major issue with that because we do expect them to go red, so it's just the appearance of the TestGrid :)

Good point which I'd not considered! I agree that it's not a major issue - hopefully it'll all get merged very quickly!

PR #779, opened 2022-12-16
Apply changes from https://github.com/cert-manager/release/pull/114

SgtCoDFish:
Should be merged after https://github.com/cert-manager/release/pull/114

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -58,8 +58,8 @@ branch-protection:
                 contexts:
                 - pull-cert-manager-master-chart
                 - pull-cert-manager-master-make-test
-                - pull-cert-manager-master-e2e-v1-25
-                - pull-cert-manager-master-e2e-v1-25-upgrade
+                - pull-cert-manager-master-e2e-v1-26
+                - pull-cert-manager-master-e2e-v1-26-upgrade
         website:
           required_status_checks:
             contexts:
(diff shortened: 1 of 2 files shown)


irbekrm (changes requested):
Thank you @SgtCoDFish 

I think we also want to update the required tests for master in this PR https://github.com/jetstack/testing/blob/master/config/config.yaml#L61-L62 apart from that all makes sense to me 👍🏼

SgtCoDFish:
> I think we also want to update the required tests for master in this PR https://github.com/jetstack/testing/blob/master/config/config.yaml#L61-L62 apart from that all makes sense to me 👍🏼

Ah you always spot useful stuff in your reviews! Of course we do, thank you!

SgtCoDFish:
I've pushed an update for the required tests. I think I should add an automated test to check for this [here](https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-presubmits.yaml) but I'll bite that off another time!

irbekrm (approved):
> I think I should add an automated test to check for this [here](https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-presubmits.yaml) but I'll bite that off another time!

Agree 👍🏼 

thanks for the update!

/lgtm

PR #780, opened 2022-12-23
use provision-buildx in tests, update OWNERS

SgtCoDFish:
Also changes resource requests

PR #781, opened 2023-01-04
Use runner for verify

SgtCoDFish:


inteon:
/approve
/lgtm
fixes problem that was introduced in https://github.com/jetstack/testing/pull/780

PR #783, opened 2023-01-11
Various config updates

SgtCoDFish:
Each commit is self-documenting. I think all of these are good to go for now - there'll be further updates needed after the release of 1.11

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -34,6 +34,9 @@ branch-protection:
       required_status_checks:
         contexts:
         - dco
+      exclude:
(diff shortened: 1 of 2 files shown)


wallrj (approved):
I'm glad you found a way to disable branch protection for dependabot.
That should allow it to rebase and force push updates, right?

/lgtm

SgtCoDFish:
>  That should allow it to rebase and force push updates, right?

I guess? I'm not really sure how it works but I'd hope so!

PR #786, opened 2023-01-27
Prevent administrators from accidentally pushing directly to protected branches

wallrj:
Causes the "Do not allow bypassing the above settings" checkbox to be ticked in the GitHub branch protection UI. See
* https://github.com/kubernetes/test-infra/blob/4afb74fe039d8d98b07397c1caba238f4bbf5fd1/prow/config/prow-config-documented.yaml#L16-L17
* https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#do-not-allow-bypassing-the-above-settings

image

One problem with this is that I expect it will prevent us fast forwarding the release branches during alpha releases and first beta release.
1. ~~The work around would be to create a PR merging master into the release branch, but that will cause a DCO problem because the merge commit will not be signed odd.~~
2. ~~Alternatively I could apply this new setting only to the master / main branches in the individual repos.~~
3. Disable branch protection temporarily during releases. This was the preferred solution when we discussed it during the standup, so I have documented it here:
  * https://github.com/cert-manager/website/pull/1173

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -31,6 +31,10 @@ branch-protection:
   orgs:
     cert-manager:
       protect: true
+      # Prevent administrators from accidentally pushing directly to protected branches.
+      # Causes the "Include Administrators" checkbox to be ticked in the GitHub branch protection UI.
+      # See https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/defining-the-mergeability-of-pull-requests/about-protected-branches#do-not-allow-bypassing-the-above-settings
+      enforce_admins: true
       required_status_checks:
         contexts:
         - dco


irbekrm (approved):
Thanks Richard this makes sense to me, I agree with the aproach of only re-enabling the force push during releases as we know that it is only needed then and it is not that often. I understand from the linked Prow docs that the branchprotector runs periodically rather than on repo settings changes, so it should not usually be the case that it would re-enable the rule before release steps are finished.

/lgtm

irbekrm:
I'll see if I can trigger the branchprotector now

irbekrm:
I'm not sure where to observe this in UI, but I triggered the branchprotector by hand and it seems to have applied the change:

```
{"client":"github","component":"branchprotector","file":"k8s.io/test-infra/prow/github/client.go:891","func":"k8s.io/test-infra/prow/github.(*client).log","level":"info","msg":"UpdateBranchProtection(cert-manager, website, jahrlin/update-release-notes-version, {\"required_status_checks\":{\"strict\":false,\"contexts\":[\"dco\",\"pull-cert-manager-website-verify\"]},\"enforce_admins\":true,\"required_pull_request_reviews\":null,\"restrictions\":null,\"required_linear_history\":false,\"allow_force_pushes\":false,\"allow_deletions\":false})","severity":"info","time":"2023-01-31T10:20:32Z"}
```

FWIW the branchprotector is a cronjob that runs every 54 minutes, so I guess any changes made to branches during release will be restored within that time period:
```
apiVersion: batch/v1
kind: CronJob
...
  schedule: 54 * * * *
  successfulJobsHistoryLimit: 3
  suspend: false
...
```

PR #788, opened 2023-02-08
Add netlify preview to the list of required checks

wallrj:
I modified the netlify configuration so that it updates the checks on cert-manager/website PRs with the status of the PR preview deployment.

image

This preview builds the site just like it is built when it is deployed to the live site so it is an ideal indication of whether the PR will work when it is merged.

This PR makes it so that the preview becomes a required check which must pass before the PR is allowed to be merged:

image

After merging this we will expect to see the following required check in all website PRs:

image

Once this is merged, I will merge:
 * https://github.com/cert-manager/website/pull/1175

....which adds a bunch of extra checks using GitHub actions, and which provides an alternative faster version of pull-cert-manager-website-verify which will run in GitHub Actions.

Then finally I will from the prow "job" with that name.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -80,6 +80,7 @@ branch-protection:
           required_status_checks:
             contexts:
             - pull-cert-manager-website-verify
+            - netlify/cert-manager-website/deploy-preview # See https://github.com/cert-manager/infrastructure#netlify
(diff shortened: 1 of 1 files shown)


inteon (approved):
Thx @wallrj

PR #791, opened 2023-02-10
Removes 1.21 tests

irbekrm:
As cert-manager v1.12 will not support Kubernetes 1.21 see https://cert-manager.io/docs/installation/supported-releases/

The change in this PR was generated from the change in cmrel https://github.com/cert-manager/release/pull/119

Note: I wonder whether it wouldn't make sense to get the cmrel command that generates Prow configs to read dynamic config params like kube version from a yaml file that we could commit to this repo. That way only one PR would be needed to do things like bump kube version.


maelvls:
> That way only one PR would be needed to do things like bump kube version.

Anything that would reduce the number of PRs needed for a simple change like this one would be very welcome!!

/lgtm

PR #794, opened 2023-02-14
Use k8s 1.26 and istio 1.15, 1.16, 1.17 for istio-csr tests

malovme:
Bump k8s version to 1.26 and istio versions to 1.12-1.16 in istio-csr presubmits

Signed-off-by: Michael Malov <14035243+malovme@users.noreply.github.com>


inteon:
@malovme thank you for the PR, will discuss tomorrow in that stand-up if everyone agrees with this change (maybe we have some kind of SLA for these older versions?)

malovme:
@inteon I've updated PR to keep only supported istio versions (https://istio.io/latest/docs/releases/supported-releases/)

PR #795, opened 2023-02-16
Remove istio v1.17 from istio-csr tests

malovme:


SgtCoDFish (approved):
/lgtm
/approve

Thanks for doing this so quickly, you're a star :grin:

PR #796, opened 2023-02-17
Various config improvements

SgtCoDFish:
This PR contains several commits each of which is documented in its commit message

### DIFF
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -203,7 +203,3 @@ plugins:
     - dco
     - owners-label
     - verify-owners
-
-  munnerz/venafi-cm-demo:
-    plugins:
-    - trigger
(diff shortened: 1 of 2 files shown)


config/plugins.yaml
```diff
@@ -157,13 +156,11 @@ plugins:
     - override
     - owners-label
     - require-matching-label
-    - shrug
```
irbekrm, line 164: > this isn't fun, but having non-functional plugins enabled presents

Agree although I was always quietly hoping to be able to use this one

irbekrm (approved):
Thanks @SgtCoDFish I think this is a great cleanup and makes the test infra more secure by removing random deps that we don't use 👍🏼 

/lgtm

PR #800, opened 2023-03-23
Bump golang-dind and add build env for golang 1.19.7

SgtCoDFish:
This follows on from #799 and creates a new golang-dind image based on the bazelbuild image created in that PR.

There'll be one more followup to use the new golang-dind image for the cmrel jobs.

```console
$ crane digest eu.gcr.io/jetstack-build-infra-images/bazelbuild:20230323-d2dfab2-4.2.3
sha256:6409016944bf6d601062062a91d283bea64834fa1f6074430d16007366a1f89c
```

PR #803, opened 2023-03-24
Bump cert-manager images with older golang-dind to newly built version

SgtCoDFish:
Followup to #800 !


irbekrm (approved):
Thanks @SgtCoDFish 

/lgtm

PR #805, opened 2023-03-28
Re-add GH required check for website

inteon:
In https://github.com/jetstack/testing/commit/c66818e542d23a43fdbf2b89f52b31be2daad51e, we incorrectly removed this check. We are still using the same check name however, now it is just a GH action.


wallrj (approved):
/lgtm

PR #806, opened 2023-03-29
Trigger make-dind build: remove comment

inteon:
This PR removes an incorrect comment.

It should also trigger the new postsubmit hook and result in the make-dind image being build.


maelvls (approved):
/lgtm

PR #808, opened 2023-03-30
Remove gcloud install, saving about 500MB from final image size

SgtCoDFish:
NB: Couldn't find any uses of gcloud or any of the other tools after a quick look in cert-manager, but it's totally possible that there's a use somewhere that I missed.

Removing python would shave off an extra 50MB but we currently still depend on that

### DIFF
--- a/images/make-dind/Dockerfile
+++ b/images/make-dind/Dockerfile
@@ -12,19 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Includes make, docker-in-docker and gcloud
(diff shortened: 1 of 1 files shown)


SgtCoDFish:
Good shout, I've updated a bunch of the comments

inteon (approved):
Thank you for the improvements!

inteon:
`verify-boilerplate` test is failing

SgtCoDFish:
My bad... our boilerplate checker in this repo is checking exactly the wrong thing going forwards. We really need to address that

PR #809, opened 2023-03-30
Switch to make-dind for our tests

inteon:
This PR switches all our tests from the bazelbuild to the make-dind image.

**I verified that `pull-cert-manager-master-make-test` works with this new image.**


wallrj (approved):
Thanks.

/lgtm

PR #810, opened 2023-03-30
Use go-dind for building images

inteon:
This PR adds a ./make-builder scripts folder (based on the builder folder) and updates all image-making tasks to use the go-dind image in combination with these ./make-builder scripts.

### DIFF
--- a/images/make-builder/ci-runner.sh
+++ b/images/make-builder/ci-runner.sh
@@ -1,0 +1,55 @@
+#!/usr/bin/env bash
+
+# Copyright 2018 The Jetstack contributors.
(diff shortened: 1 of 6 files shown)


inteon:
works in my local environment

images/make-builder/main.go
inteon, line 1: This file is copied from images/builder/main.go

images/make-builder/ci-runner.sh
inteon, line 1: Based on images/builder/ci-runner.sh but without the upgrade scripts (that is for a later PR).

PR #811, opened 2023-03-30
Reduce size of go-dind image

inteon:
This change should reduce the size of the go-dind image.


SgtCoDFish (approved):
/lgtm
/approve

Haha I think I spotted this a while ago, meant to do something about it and then forgot about it! Nice catch, thanks!

PR #813, opened 2023-03-30
Cleanup presubmits

inteon:
This PR removes some default values and old docker-related values.


wallrj (approved):
I don't know what `context` or `agent` fields are, but I trust you've looked them up and have confidence that they are not needed.

Try it and see I guess.

/lgtm

inteon:
> context

The values that we provided equal the default values for these fields.

PR #815, opened 2023-03-31
Rename yaml to match upstream

inteon:


irbekrm:
@inteon I normally apply these changes and verify that it works and only then create a PR afaik this is also how @munnerz used to do that

inteon:
@irbekrm this should be ready to be merged now. The cluster has been updated with the changes.

irbekrm (approved):
Thanks @inteon 

/lgtm

PR #816, opened 2023-03-31
Rename cherrypick to cherrypicker in external_plugins

inteon:
this was missing from #815

Since we rename the deployment and service from `cherrypick` to `cherrypicker`, we also have to rename `external_plugins`.


irbekrm (approved):
/lgtm

PR #818, opened 2023-04-03
Deprecate jetstack projects

inteon:
This PR removes references & tests for all projects in the jetstack organisation.
This is a list of all affected repos:
- jetstack/kube-lego [Public archive]
- jetstack/puppet-module-aws_ebs [Public archive]
- jetstack/puppet-module-calico [Public archive]
- jetstack/puppet-module-etcd [Public archive]
- jetstack/puppet-module-kubernetes [Public archive]
- jetstack/puppet-module-kubernetes_addons [Public archive]
- jetstack/puppet-module-prometheus [Public archive]
- jetstack/puppet-module-tarmak [Public archive]
- jetstack/puppet-module-vault_client [Public archive]
- jetstack/tarmak [Public archive]
- jetstack/test-infra [Public archive]
- jetstack/vault-plugin-auth-templated-aws [Public archive]
- jetstack/vault-helper [Public archive]
- jetstack/vault-unsealer [Public archive]
- jetstack/navigator [Public archive]
- jetstack/okta-kubectl-auth [Public archive]
---
- jetstack/version-checker [TODO: should be GH-only/ Public archive] -> I propose we Public archive
- jetstack/kube-oidc-proxy [TODO: should be GH-only/ Public archive] -> I propose we Public archive
---
- jetstack/cert-manager-nginx-plus-lab [OK: the repo was only using "label_sync"]
- flightdeck / katacoda [OK: katacoda is EOL]

If any of these repos are still used, we should migrate that repo to use GH actions (please add a note in the comments).
Ideally, we also publically archive all repos that are not expected to receive any updates in the future.

PR #819, opened 2023-04-04
Remove old image builds

inteon:
In #818, we removed all private jetstack testing configs.
This PR now also removes the leftover image build configs for these repos.


SgtCoDFish (approved):
/lgtm
/approve

Makes sense and - again - easy to roll back! Thanks!

PR #820, opened 2023-04-04
Replace bazelbuild with make-dind (test)

inteon:
This PR switches one test from the bazelbuild to the make-dind image.

We do not need the `runner` argument, because this script is already the entrypoint.

### DIFF
--- a/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
+++ b/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
@@ -18,9 +18,8 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20230323-d2dfab2-4.2.3
+      - image: eu.gcr.io/jetstack-build-infra-images/make-dind:20230330-83ce2fd-bullseye
         args:
-        - runner
         - make
         - -j2
         - vendor-go


config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
```diff
@@ -18,9 +18,8 @@ presubmits:
       preset-service-account: "true"
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/bazelbuild:20230323-d2dfab2-4.2.3
+      - image: eu.gcr.io/jetstack-build-infra-images/make-dind:latest-bullseye
```
wallrj, line 21: I doubt whether @irbekrm  or @SgtCoDFish would agree with using a `latest-` tag here.
It should be a sha sum, to avoid any chance that someone accidentally (or deliberately) publishes a rogue image with that tag.

Previously it was at least a somewhat unique tag, which would avoid instantly breaking all the tests by publishing a new (broken) build image.

inteon: The image only contains make and docker, it should not be possible to break our tests by upgrading the image.

wallrj: I guess what I want to know is why have you changed from a very specific tag to a generic latest tag?
Are you trying to simplify the process of upgrading the build image...avoiding the need to update all the jobs when it gets upgraded?
Are you intending to regularly rebuild the `latest-bullseye` image based on an `apt upgrade`?
I guess it's not such a risk as long as  we trust that Debian won't break things within the bullseye release.
The date specific tags do still exist:

```
$ gcrane ls eu.gcr.io/jetstack-build-infra-images/make-dind

eu.gcr.io/jetstack-build-infra-images/make-dind@sha256:5b13ab93fe6535a9f0277370c34cafc1e7e8540ac4ae34eba991a7a16745abdd
eu.gcr.io/jetstack-build-infra-images/make-dind:20230329-e9d7eb3-bullseye
eu.gcr.io/jetstack-build-infra-images/make-dind@sha256:c8b46f4d9bca8dd7be8a02b1557f1f247fb9ba518cd2dd71002f3a295296bae2
eu.gcr.io/jetstack-build-infra-images/make-dind:20230330-83ce2fd-bullseye
eu.gcr.io/jetstack-build-infra-images/make-dind:latest-bullseye
```

> The image only contains make and docker, it should not be possible to break our tests by upgrading the image.

I mean its certainly possible. 
Which version of `make`, which version of `docker`?
You could easily break things by upgrading or downgrading either of these.
 * E.g. see breaking changes in https://lists.gnu.org/archive/html/info-gnu/2023-02/msg00011.html

Is it the intention that only those two executables are present in the image?
Because there are hundreds of executables in there, many of which might be shelled out to by the cert-manager tests.

```bash
docker run -it --rm eu.gcr.io/jetstack-build-infra-images/make-dind:latest-bullseye ls /bin /usr/bin /usr/local/bin
```

Is it possible / simple to rebuild the `make-dind` image without updating the `latest-bullseye` tag?
Otherwise how would I experimentally build a new image where I upgrade to a newer version of Docker,
without instantly having all the tests use that image.

I'm waffling. I don't agree with the use of `latest-bullseye` but if that's what you've planned with @SgtCoDFish I'm happy to give way.

/lgtm
/hold in case others object to this.

inteon: The advantage of using latest: you don't have to change the tag to upgrade to the latest versions & fix CVEs.
The disadvantage: it's still possible to break tests in case of breaking changes.

@wallrj I think you are correct, I underplayed the chances of bumping into breaking changes. Also, for LTS there will be a lot more breaking changes (since that will be a longer period of time).

wallrj (approved):
Thanks.

/lgtm
/unhold

PR #822, opened 2023-04-05
Use updated make-dind image that includes jq

inteon:


wallrj (approved):
/lgtm

This is why I may have been wrong about using the unique build image tags.

PR #827, opened 2023-04-05
Add logic for local cache in make-dind

inteon:
In this PR, the runner script tries to prevent any concurrency caching issues by only using atomic operations when working on shared files & folders.

The main purpose of this caching solution is to allow concurrent caching: if one test is still downloading file A, we don't want another test to think that file is ready already and start executing a half-downloaded file.

For this reason we are taking the following approach:
1. on startup, we look for the location of the latest finished cache in the "${SHARED_CACHE_DIR}/latest" file
2. we copy the contents of this cache folder to our local cache
3. ... running test ...
4. on successful completion, we move our local cache to a newly created unique folder in the shared cache mount
5. after the cache is successfully moved, we update the "${SHARED_CACHE_DIR}/latest" file with the location of this moved cache
6. this moved cache won't be altered, so it is safe to be read from multiple concurrent tests
7. at some point the "${SHARED_CACHE_DIR}/latest" value will point to another cache
8. some tests might still be copying (step 2) an older cache from a location that differs from "${SHARED_CACHE_DIR}/latest"
9. to make sure we don't break any of those copies, we keep the 4 newest caches and remove all the older caches to limit disk usage

### DIFF
--- a/images/make-dind/runner
+++ b/images/make-dind/runner
@@ -16,6 +16,37 @@
 
 # generic runner script, handles DIND, etc.
 
+
+# Check if the job has opted-in to local caching and if so check for a cache and
+# copy the cache to prepopulate the local cache. After the job is done, change the
+# latest cache directory to the local cache directory.
+export LOCAL_CACHE_ENABLED=${LOCAL_CACHE_ENABLED:-false}
+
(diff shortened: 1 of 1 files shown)


maelvls:
This seems complicated. What is the purpose?

PR #829, opened 2023-04-06
Use upstream Makefiles and scripts instead of bazel

inteon:


irbekrm:
> Use upstream Makefiles and scripts instead of bazel

Can you please link where they come from?

prow/Makefile
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+.PHONY: update-config
+update-config:
+	kubectl create configmap config --from-file=config.yaml=config.yaml --dry-run=client -o yaml | kubectl replace configmap config -f -
+
+.PHONY: update-plugins
+update-plugins:
+	kubectl create configmap plugins --from-file=plugins.yaml=plugins.yaml --dry-run=client -o yaml | kubectl replace configmap plugins -f -
+
```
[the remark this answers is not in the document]
inteon: I assume this is handy when you want to test a config change without using git?
Do I add a comment or do I remove these targets?

prow/pj-on-kind.sh
```diff
@@ -13,117 +15,15 @@
...
-# */
-
-# This script is copied from k/test-infra
-# https://github.com/kubernetes/test-infra/blob/488e767e326f6c7189cbf0682e7f926040ae959c/prow/pj-on-kind.sh
-# Runs prow/pj-on-kind.sh with config arguments specific to Jetstack Prow config.
-# Requries go, docker, and kubectl.
 
-# Example usage:
-# ./prow/pj-on-kind.sh ci-cert-manager-e2e-v1-21
+# Runs prow/pj-on-kind.sh with config arguments specific to the prow.k8s.io instance.
```
irbekrm, line 22: I don't think we want to remove this comment or add a reference to `prow.k8s.io` which is not what we are targeting

prow/pj-on-kind.sh
```diff
@@ -13,117 +15,15 @@
...
-      rm "${temp_config}"
-    fi
-  fi
-}
+root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd -P)"
+export CONFIG_PATH="${root}/config/config.yaml"
+export JOB_CONFIG_PATH="${root}/config/jobs"
 
-main "$@"
+bash <(curl -s https://raw.githubusercontent.com/kubernetes/test-infra/master/prow/pj-on-kind.sh) "$@"
```
irbekrm, line 31: I would rather keep this file checked into the codebase than pull in a random script from upstream

irbekrm: FWIW we had a discussion about this already https://github.com/jetstack/testing/pull/526#discussion_r667956641

prow/Makefile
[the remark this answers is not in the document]
irbekrm: tbh, it is going to be inconsistent in either way. My main concern is that folks will think that this is how config changes should be applied. A warning that this is just for local testing and any applied changes should be either reset or pushed to Git shoudl be ok 👍🏼
 and perhaps in future we can make it so that the way Prow and Prow config gets applied is more consistent instead of one from dev laptop, the other from Git

prow/cluster/labelsync_cronjob.yaml
```diff
@@ -28,7 +28,7 @@ spec:
         spec:
           containers:
             - name: label-sync
-              image: eu.gcr.io/jetstack-build-infra/prow-labelsync:latest
+              image: gcr.io/k8s-prow/label_sync:v20230405-89944e7c3e
```
irbekrm, line 31: There is probably a huge diff here as the last time labelsync was pushed to our registry was in 2019 https://console.cloud.google.com/gcr/images/jetstack-build-infra/eu/prow-labelsync?authuser=1&project=jetstack-build-infra
Please take a look at the release notes and add a note whether there is any diff in how it works that would affect us.

prow/cluster/mem-limit-range_limitrange.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: v1
-kind: LimitRange
```
irbekrm, line 2: Why do we remove this?

inteon: Because it is yaml for the worker cluster/ is not deployed on the prow cluster. Maybe I should place it in a separate dir?

irbekrm: No, that's my bad - and I see that the workers cluster anyway does not have any limitrange in test-pods namespace, so should be fine to remove

prow/cluster/labelsync_cronjob.yaml
inteon: @irbekrm where do I find these release notes? Do I manually check the git diff?

irbekrm: I normally just skim through Prow announcements, I see that this is now moved to https://docs.prow.k8s.io/docs/announcements/

prow/Makefile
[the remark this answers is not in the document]
inteon: done

irbekrm (approved):
Thank you for making these improvements! Adding an lgtm with a hold for you to merge when ready

/lgtm
/hold

inteon:
I applied the changes, and it seems to be working still.
/unhold

PR #830, opened 2023-04-06
BUGFIX: Replace make-dind with golang-dind where replaced incorrectly

inteon:
In https://github.com/jetstack/testing/pull/809, we switched from bazelbuild to make-dind.
However, in that PR, some golang-dind images were incorrectly replaced with make-dind.
This PR fixes that mistake and replaces those golang-dind images with golang images based on the latests make-dind version.


wallrj (approved):
/lgtm

PR #831, opened 2023-04-06
Add GOCACHE to cert-manager tests

inteon:
Adds a preset for enabling a persistent go cache. This cache is a shared folder on the host machine, concurrent use of the GOCACHE is allowed, so we can just point all tests to the same cache.


maelvls:
Let's monitor how Prow reacts and see whether we see somethihng like https://github.com/jetstack/testing/pull/720 when we had to disable Go caching.

/lgtm

PR #832, opened 2023-04-06
Remove unused image bazel-tools

inteon:
This image was added in https://github.com/jetstack/testing/pull/518. However, this image is not used anymore for building FIPS images. For this reason, it can be removed.


maelvls (approved):
Thanks for cleaning this up.

/lgtm

PR #848, opened 2023-04-07
Remove all remaining bazel and OWNERS files

inteon:
- removes old OWNERS files
- removes old bazel files
- changes testing presubmit to make instead of bazel
- removes old bazel presets


maelvls:
All good, thanks.

PR #849, opened 2023-04-07
Update README, add local tests and rename builder script

inteon:
Fixes some remaining TODOs from https://github.com/jetstack/testing/pull/848.


images/image-builder-script/builder.sh
```diff
@@ -42,7 +42,7 @@ fi
 
 echo "Executing builder..."
 PUSHED_IMAGE=$(cd "$SCRIPT_DIR" && \
-    go run . --build-dir "${PROJECT_DIR}"/"${BUILD_DIR}" "$@")
+    go run . --build-dir "${PROJECT_DIR}"/"${BUILD_DIR}" $@)
```
maelvls, line 45: hmm i think you want to keep the double quotes here, otherwise the arguments with spaces will get treated as different arguments

example:

```
images/image-builder-script/builder.sh "foo bar"
```

will be interpreted as

```
go run . --build-dir "/home/mvalais/code/cert-manager/_build" foo bar
```

instead of

```bash
go run . --build-dir "/home/mvalais/code/cert-manager/_build" "foo bar"
```

```suggestion
    go run . --build-dir "${PROJECT_DIR}"/"${BUILD_DIR}" "$@")
```

inteon: but what happens when the provided value is eg. "--test1 value1 --test2 value2"

maelvls:
Thanks for updating the README to reflect the fact that Bazel is gone.

I left a tiny nit

/lgtm

maelvls:
unhold if you think the nit is a nit

/hold

images/image-builder-script/builder.sh
inteon: Thanks, this seems to work!

PR #853, opened 2023-04-08
Update config.yaml and plugins.yaml

inteon:
fixes wrong links, deprecated options and missing upstream options


config/config.yaml
```diff
@@ -17,12 +17,25 @@ plank:
...
+            cpu: 100m
+        initupload:
+          requests:
+            cpu: 100m
+        place_entrypoint:
+          requests:
+            cpu: 100m
+        sidecar:
+          requests:
+            cpu: 100m
```
irbekrm, line 38: Are these copied from upstream or are on basis of how much resources we need? And how does it compare with what we're requesting now?

inteon: Currently, there are no requests for those init containers:
from the pod's yaml:
```yaml
resources: {}
```

These requests were copied from upstream: https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L26-L38

config/config.yaml
```diff
@@ -103,29 +116,39 @@ sinker:
 
 deck:
   spyglass:
-    size_limit: 500000000 # 500MB
+    size_limit: 100000000 # 100MB
```
irbekrm, line 119: Has this change been tested?

inteon: Not yet, I've been using the git<->prow config sync to make changes to the config.
So I test by commit to this repo.

PR #869, opened 2023-04-11
Remove gs:// prefix from config

inteon:
Was introduced as part of #853 based on comments in the config.
After testing, it turns out that this breaks our testgrid links, these link to broken links eg. https://prow.build-infra.jetstack.net/view/gs/gs:/jetstack-logs/logs/ci-cert-manager-master-make-test/1645782628332212224.


wallrj:
Self reviewed and merged?

inteon:
> Self reviewed and merged?

Indeed.
For context, this is a small bugfix. It reverts a change that was made in https://github.com/jetstack/testing/pull/853.
It would have been very useless to get a review for this PR.

PR #870, opened 2023-04-12
Double requested CPU for our e2e tests based on observed usage

inteon:
image

Due to our changes in https://github.com/cert-manager/cert-manager/pull/5932, the e2e tests use a lot more CPU.
We have to adapt the amount of requested CPU for these tests so we don't over-schedule on a node.


SgtCoDFish (approved):
/lgtm
/approve

OK as a quick test, but should be done in cmrel later!

PR #873, opened 2023-04-13
Rename trusted cluster files and move commenter periodics to the trusted cluster

inteon:


irbekrm (approved):
Thanks @inteon this makes sense.
I checked the TestGrid config generator job, but it appears that it should still be able to generate TestGrid config for these jobs even when they are in trusted cluster as we read all job configs https://github.com/jetstack/testing/blob/master/config/jobs/testing/testing-trusted.yaml#L32

Once this merges, we should double check that the jobs run as expected and the results appear in TestGrid https://testgrid.k8s.io/jetstack-testing-janitors

/lgtm

PR #874, opened 2023-04-13
Add DCO exception for jetstack-bot

inteon:
@maelvls This should solve the problem we have in https://github.com/jetstack/testing/pull/872.

### DIFF
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -91,6 +91,10 @@ require_matching_label:
   prs: true
   regexp: ^kind/
 
+dco:
+  jetstack:
+    trusted_apps: [ "jetstack-bot" ]
+
 plugins:
 
   jetstack:


config/plugins.yaml
```diff
@@ -91,6 +91,10 @@ require_matching_label:
   prs: true
   regexp: ^kind/
 
+dco:
+  jetstack:
+    trusted_apps: [ "jetstack-bot" ]
```
wallrj, line 96: The documentation talks about not including the `bot` suffix....although I don't know how it would assume that all Bot accounts have that suffix:

https://github.com/kubernetes/test-infra/blob/a98f7f6910df237202851f6ab0b8b43dec2859b5/prow/plugins/config.go#L682-L690

> TrustedApps defines list of apps which commits will not be checked for DCO singoff.
> The list should contain usernames of each GitHub App without [bot] suffix.
> By default, this option is ignored.
>

inteon: @wallrj could you approve the PR? We can only be 100% certain by testing this configuration 😄 .

wallrj (approved):
/lgtm

PR #875, opened 2023-04-13
Use prowjob to setup GH webhooks + hmac secret

inteon:
see: https://docs.prow.k8s.io/docs/components/optional/hmac/

Currently, it seems like the GH webhooks are configured manually.

GH uses a HMAC secret to sign events that are sent to the prow webhook.
This prowjob should automatically generate the HMAC secret and create the webhooks in GH.
The HMAC secret should be created in the "prowjob" cluster, this prowjob is created in the "trusted" cluster however (so we still have to create a kubeconfig secret).
Configuring this setup here will make it easier to rotate HMAC secrets & create new clusters.

Plan of action:
- create a secret in the "prowjob" cluster that contains the kubeconfig of the "trusted" cluster
- merge & run the prowjob
- set `dry-run=false`, which should result in the `hmac-token-new` secret being created
- try using the new `hmac-token-new` secret
- remove the old webhook configs in GH

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -37,6 +37,17 @@ plank:
           requests:
             cpu: 100m
 
+managed_webhooks:
+  # This has to be true if any of the managed repo/org is using the legacy global token that is manually created.
+  respect_legacy_global_token: true
+  
+  # Config for orgs and repos that have been onboarded to this Prow instance.
+  org_repo_config:
+    jetstack/testing:
+      token_created_after: 2023-04-13T00:00:00Z
+    cert-manager:
+      token_created_after: 2023-04-13T00:00:00Z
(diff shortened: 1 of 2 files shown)


irbekrm:
I think this needs a lot more documentation and a proper PR description.

Can you please
- add to the PR description what will actually change in our cluster as a result of merging this PR given that we have already onboarded these two repos - is it safe to make that change now?
- add to the PR description why this is better than what we have now- what's the difference
- add to the Readme how the repo access is set up and what actions a user would perform, what are these secrets, how will they be generated etc
- clearly write up what follow up work you plan to do for this

inteon:
closing since this is probably not required.

PR #879, opened 2023-04-24
Add prowgen to our CI

inteon:
This PR adds prowgen to our CI, so it can never be out-of-sync.


Makefile
```diff
@@ -16,8 +16,22 @@
...
 
+.PHONY: prowgen
+prowgen:
+	cd ./config/prowgen/ && go run . --branch=* -o $(CURDIR)/config/jobs/cert-manager/cert-manager/
+
+.PHONY: verify-prowgen
+verify-prowgen:
+	mkdir -p _temp
+	cd ./config/prowgen/ && go run . --branch=* -o $(CURDIR)/_temp/cert-manager/
+	diff -q -r $(CURDIR)/config/jobs/cert-manager/cert-manager/ $(CURDIR)/_temp/cert-manager/
```
irbekrm, line 27: @inteon I am not sure about this - for example if we run the check now it will require that the changes such as the CPU request bump are applied to release-1.10 tests where they are not actually needed. I think before we can have a check like this any changes should be configurable for a specific branch only

Makefile
```diff
@@ -16,8 +16,22 @@
 verify-boilerplate:
 	@./hack/verify-boilerplate.py --rootdir=$(CURDIR) --boilerplate-dir=hack/boilerplate && echo "Boilerplate verification passed."
 
+.PHONY: prowgen
+prowgen:
+	cd ./config/prowgen/ && go run . --branch=* -o $(CURDIR)/config/jobs/cert-manager/cert-manager/
```
irbekrm, line 21: It'd be nice to have the branch configurable as I think a common pattern when making changes will be to apply them to master branch only

inteon: I think that all branches should always be in sync with the code, so I'm not sure why you would want to update once branch only. If needed for local development only, you can run the command yourself or undo the changes in git.

config/prowgen/main.go
```diff
@@ -0,0 +1,183 @@
...
+	f, err := os.Create(path)
+	if err != nil {
+		return err
+	}
+
+	if _, err := io.Copy(f, strings.NewReader(data)); err != nil {
+		return err
+	}
+
+	return nil
```
irbekrm, line 174: I prefer to be able to output to stdout, inspect it and then, if I want to, redirect it to a file when I am making some changes by hand rather than having to figure out how to pass a path to a file and read it from there. I think we should leave the stdout option in - any templating tool I can think of can write to stdout.

Could we perhaps only make changes that are required to move these files in this PR and not additional unrelated changes?

inteon: on line 152 we still have `if o.OutputDir == "" {` which allows us to output to console.
The reason I did change this logic is that I wanted to be able to specify where to output the files (if not via stdout).

inteon: @irbekrm I'll create a seperate PR that just copies the code, so this PR becomes easier to review.

inteon: @irbekrm  https://github.com/jetstack/testing/pull/880

SgtCoDFish (approved):
/lgtm
/approve

Thanks, looks great!

PR #880, opened 2023-04-24
Copy prowgen from cert-manager/release

inteon:
This is the first step in the process of moving prowgen to cert-manager/release. The only changes that were made are filename changes & module changes.

The copy is based on https://github.com/cert-manager/release/commit/49b0c286b27ae0a37d656136afe18de2771318bd

### DIFF
--- a/config/prowgen/go.mod
+++ b/config/prowgen/go.mod
@@ -1,0 +1,11 @@
+module prowgen
+
+go 1.20
+
+require (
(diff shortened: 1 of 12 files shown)


irbekrm (approved):
Thanks Tim, I think moving the prow job generation here will make things a lot easier 👍🏼 

I've just read through the files and compared the contents, haven't tried running it from here yet.

/lgtm

PR #882, opened 2023-04-24
Setup testing repo presubmits so prowgen is tested & go is available for the verify target

inteon:
This PR runs the go tests before merging testing PRs.
Also, it prepares the make verify target with a golang installation so go can be used in a follow-up PR.

preparation for #879


irbekrm (approved):
Thanks Tim

/lgtm

PR #885, opened 2023-04-26
Document steps necessary to setup worker cluster

inteon:
Together with @irbekrm, I fixed the crier kubeconfig setup. At the same time (while discovering what is configured on our worker clusters), I wrote down the steps necessary to configure a worker cluster.

### DIFF
--- a/prow/_hack/create_user.sh
+++ b/prow/_hack/create_user.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# Change USERNAME to your username (e.g. client or crierclient)
+USERNAME=crierclient
+
+CSR_FILE=$USERNAME.csr
+KEY_FILE=$USERNAME.key
+CRT_FILE=$USERNAME.crt
+
+openssl genrsa -out $KEY_FILE 2048
+openssl req -new -key $KEY_FILE -out $CSR_FILE -subj "/CN=$USERNAME"
+
+cat <<EOF | kubectl create -f -
+apiVersion: certificates.k8s.io/v1
+kind: CertificateSigningRequest
+metadata:
+    name: $USERNAME
+spec:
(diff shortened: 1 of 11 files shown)


prow/_hack/create_user.sh
```diff
@@ -0,0 +1,44 @@
...
+cat <<EOF | kubectl create -f -
+apiVersion: certificates.k8s.io/v1
+kind: CertificateSigningRequest
+metadata:
+    name: $USERNAME
+spec:
+    signerName: kubernetes.io/kube-apiserver-client
+    groups:
+    - system:authenticated
+    request: $(cat $CSR_FILE | base64 | tr -d '\n')
```
SgtCoDFish, line 36: suggestion (non-blocking): This kind of thing gives me the heebie-jeebies. It's so easy for a command to result in invalid YAML, or - worse - YAML which has a different meaning to what was intended. If that isn't possible today, it could easily happen in a future version where this command subtly changes. Unix pipes deal with text and YAML isn't regular text.

I'd vote to use `yq` instead to be sure that the variable is substituted properly (or use `jq` and pass in a JSON blob). Since this is a script which won't be used often, I guess it doesn't matter a great deal and this isn't a blocker!

SgtCoDFish (approved):
/lgtm
/approve
/hold

Hold because I have a non-blocking suggestion, but this is a big improvement as-is and I'd happily merge - thanks :grin:

prow/_hack/create_user.sh
inteon: Thanks for the suggestion, I generally agree, however I would like to not depend on yq/ jq. This case seems somewhat safe-ish, it is a one-line base64 blob, but I agree that this can go wrong pretty quickly.

PR #887, opened 2023-05-09
Bumps Go

irbekrm:
Bumps Go to 1.20 for all images used for tests of projects in cert-manager org.

It appears that kube 1.27 libs require Go 1.20 as a minimum, so we cannot test upgrades with 1.19 images, see i.e https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/pr-logs/pull/cert-manager_webhook-example/50/pull-cert-manager-webhook-example-verify/1655980875042525185


SgtCoDFish (approved):
/lgtm
/approve

Thanks :+1:

PR #888, opened 2023-05-11
Add issuer-lib repo

inteon:
Adds the issuer-lib repo to our prow setup.
The repo is still private, I'll make it public once it has been configured correctly by prow (after this PR is merged).

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -103,6 +103,12 @@ branch-protection:
             contexts:
             - pull-trust-manager-verify
             - pull-trust-manager-smoke
+        issuer-lib:
+          required_status_checks:
+            contexts:
+            - lint
+            - test-e2e
+            - test-unit
         csi-lib:
           required_status_checks:
             contexts:
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -10,6 +10,7 @@ triggers:
   - cert-manager/cert-manager
   - cert-manager/website
   - cert-manager/trust-manager
+  - cert-manager/issuer-lib
   only_org_members: true
 
 blunderbuss:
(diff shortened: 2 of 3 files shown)


config/config.yaml
```diff
@@ -103,6 +103,12 @@ branch-protection:
             contexts:
             - pull-trust-manager-verify
             - pull-trust-manager-smoke
+        issuer-lib:
+          required_status_checks:
+            contexts:
+            - lint
+            - test-e2e
+            - test-unit
```
irbekrm, line 111: where is it going to find those tests?

inteon: those are GH action tests.

inteon: eg. https://github.com/cert-manager/issuer-lib/pull/7/files#diff-da26bf1b5f45a27fe78427b79c8711292263ae8518581763715d09723126dfabR5

maelvls:
Could we have some form of notice to the other maintainers to let them know that Jetstack/Venafi wants to donate issuer-lib?

A cheap way of doing that would be to share that during the biweekly meeting. I don't think anyone will disagree. I think it is worth forcing ourselves (Jetstack employees) to be transparent towards other maintainers.

(I don't think this is blocking)

inteon:
> Could we have some form of notice to the other maintainers to let them know that Jetstack/Venafi wants to donate issuer-lib?
> 
> A cheap way of doing that would be to share that during the biweekly meeting. I don't think anyone will disagree. I think it is worth forcing ourselves (Jetstack employees) to be transparent towards other maintainers.
> 
> Hopefully we will have some form of steering committee to streamline this kind of process.
> 
> (I don't think this is blocking)

I added this item to the agenda of our next biweekly meeting. Thanks for noting.

PR #890, opened 2023-05-23
Use Recreate as deployment strategy for ghproxy

inteon:
Since the PVC can only be mounted in 1 pod, the rolling update strategy resulted in a second pod that was not able to startup.
The `Recreate` strategy fixes this issue.


irbekrm (approved):
Thanks @inteon this makes sense, I see that upstream does that as well https://github.com/kubernetes/k8s.io/blob/main/apps/prow/cluster/ghproxy_deployment.yaml#L45-L46

/lgtm

PR #891, opened 2023-05-23
Upgrade prowjob CRD

inteon:
Upgrades the prowjob CRD, so it matches upstream ( matches the image upgrades in #872 ).


inteon:
Applied on cluster.

PR #897, opened 2023-08-25
Change master's primaryKubernetesVersion to 1.28

inteon:


wallrj (approved):
Ok. So we are dropping support for 1.22? 

The documentation still says that 1.13 will support 1.22 -> 1.27
 * https://cert-manager.io/docs/installation/supported-releases/#upcoming-releases

But looks like previous releases also bumped the minimum  version too.

We need to update the https://cert-manager.io/docs/installation/supported-releases/#kubernetes-supported-versions section with the latest supported cloud versions.

/lgtm

PR #899, opened 2023-08-29
Disable automatic self-approval for PRs

SgtCoDFish:
This was discussed in standup on 29/08/2023. This relates to https://github.com/cert-manager/cert-manager/pull/6260 and our efforts to define different roles within the org.

Since any org member can LGTM, having this enabled means that any approver (today, basically the maintainers) who raises a PR can have their PR merged by anyone else in the org dropping an `/lgtm`.

Instead, we ideally want a _different_ approver to have to approve a PR while encouraging drive-by `/lgtm`s from stakeholders. This still allows approvers to self-approve, which might be appropriate for e.g. fixing typos in READMEs or other trivial tasks, but shouldn't be the default. There doesn't seem to be an obvious way to disable the ability for these auto-approvals.

### DIFF
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -93,6 +93,10 @@ require_matching_label:
   prs: true
   regexp: ^kind/
 
+approve:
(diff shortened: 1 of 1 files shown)


inteon (approved):
/approve
/lgtm

PR #900, opened 2023-08-29
Only allow OWNERS to /lgtm

inteon:
https://github.com/kubernetes/test-infra/blob/324168cf409e3e2ff2413b7114240c7d2f95d8db/prow/plugins/config.go#L192-L195C36


SgtCoDFish (approved):
/lgtm
/approve

Aaaaah, good spot! 😁

PR #902, opened 2023-08-30
Add trust-manager to the --only flag

inteon:
I already applied this change, and the labels are synced now.


SgtCoDFish (approved):
/lgtm
/approve

Phew I'd have not thought to look there!

Thanks for finding this and fixing it 😁

PR #909, opened 2023-09-19
Apply upstream changes

inteon:
Update prow yaml based on upstream (result of manually comparing yaml with upstream).
see https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster

### DIFF
--- a/prow/cluster/crier_deployment.yaml
+++ b/prow/cluster/crier_deployment.yaml
@@ -40,7 +40,7 @@ spec:
         - --github-endpoint=http://ghproxy
         - --github-endpoint=https://api.github.com
         - --github-token-path=/etc/github/oauth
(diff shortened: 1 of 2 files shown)


inteon:
I applied these changes to the cluster & everything is still working.

maelvls:
What do you mean by upstream? I am a bit confused by where these changes come from

inteon:
> What do you mean by upstream? I am a bit confused by where these changes come from

I got the changes from here: https://github.com/kubernetes/test-infra/tree/master/config/prow/cluster

SgtCoDFish (approved):
/lgtm
/approve

Obviously I can't really review the diff on the CRD given the size. I think we should proceed and look later if there's a problem 😁

PR #910, opened 2023-09-25
Stop testing cert-manager 1.11 (because it is EOL)

inteon:
see https://cert-manager.io/docs/installation/supported-releases/


SgtCoDFish (approved):
/lgtm
/approve

Thanks for picking this up!

PR #913, opened 2023-10-20
Switch nodeport services to clusterIP

inteon:
I'm don't know why Kubernetes makes all these services NodePort services, but I don't think we need it.

I applied this config on 20/10/2023 and will report whether we experience any problems.


inteon:
Seems to be working 100% correctly still.

SgtCoDFish (approved):
/lgtm
/approve

Trusting you that it works correctly 😁

PR #914, opened 2023-10-20
Fix error in autobumper configuration

inteon:
This error is caused by our migration from jetstack/testing to cert-manager/testing.
We forgot to update this field in the autobump config.


SgtCoDFish (approved):
/lgtm
/approve

PR #920, opened 2023-11-05
chore(approver-policy): replace golang image with jetstack:golang-dind

erikgb:
To install `protoc` in https://github.com/cert-manager/approver-policy/pull/277 I need `unzip` available in the CI images. One of the approver-policy CI-jobs (the single one not failing on the PR) already uses the Jetstack golang image, so I think this change can be justified.

/cc @irbekrm


irbekrm (approved):
Thanks Erik!

PR #921, opened 2023-11-14
Limit dind mtu to 1460

inteon:
Our docker-in-docker setup was incorrectly using a 1500 mtu which is larger than the 1460 mtu of the eth0 device that is used in the GKE pod.
This causes hard-to-debug networking issues.

before:
```
root@test-pod:/workspace# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
3: eth0@if36: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1460 qdisc noqueue state UP group default
4: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
6: vethd1f289b@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master docker0 state UP group default
```

after:
```
root@test-pod:/workspace# ip addr
1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
3: eth0@if36: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1460 qdisc noqueue state UP group default
4: docker0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1460 qdisc noqueue state UP group default
6: vethd1f289b@if5: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1460 qdisc noqueue master docker0 state UP group default
```

see https://blog.zespre.com/dind-mtu-size-matters.html for more info

### DIFF
--- a/images/make-dind/Dockerfile
+++ b/images/make-dind/Dockerfile
@@ -73,7 +73,9 @@ RUN apt-get update \
     && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Move Docker's storage location & enable experimental features & add Container Registry cache (see: https://cloud.google.com/container-registry/docs/pulling-cached-images)
-RUN echo 'DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph --experimental --registry-mirror=https://mirror.gcr.io"' | \
+# @inteon: added --mtu 1460 to fix network issues due to parent mtu < child mtu (see https://blog.zespre.com/dind-mtu-size-matters.html)
+#          at the time of writing, the parent mtu is 1460 (see https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview)
+RUN echo 'DOCKER_OPTS="${DOCKER_OPTS} --mtu=1460 --data-root=/docker-graph --experimental --registry-mirror=https://mirror.gcr.io"' | \
     tee --append /etc/default/docker
 
 # NOTE this should be mounted and persisted as a volume ideally (!)


images/make-dind/Dockerfile
```diff
@@ -73,7 +73,8 @@ RUN apt-get update \
     && update-alternatives --set ip6tables /usr/sbin/ip6tables-legacy
 
 # Move Docker's storage location & enable experimental features & add Container Registry cache (see: https://cloud.google.com/container-registry/docs/pulling-cached-images)
-RUN echo 'DOCKER_OPTS="${DOCKER_OPTS} --data-root=/docker-graph --experimental --registry-mirror=https://mirror.gcr.io"' | \
+# @inteon: added --mtu 1400 to fix network issues due to parent mtu < child mtu (see https://blog.zespre.com/dind-mtu-size-matters.html)
+RUN echo 'DOCKER_OPTS="${DOCKER_OPTS} --mtu=1400 --data-root=/docker-graph --experimental --registry-mirror=https://mirror.gcr.io"' | \
```
SgtCoDFish, line 77: suggestion: Is it worth adding in this comment what the parent MTU was at the time the flag was added, which would help explain to people reading the code why the value of the MTU here was chosen? This comment explains why we set the MTU but doesn't explain the value we chose.

question: Why 1400? If the parent is 1460 we're potentially wasting a bunch of packets by setting the MTU to 1400 here. 1460 should be safe while being more efficient, no?

SgtCoDFish (commented):
This looks super interesting - how did you come to suspect the MTU?

Got a suggestion / question.

images/make-dind/Dockerfile
inteon: I randomly picked 1400 so the fix would still work if the parent mtu changed to eg. 1450.
I'll already add a note stating that the mtu was 1460 at the time of writing this fix.

inteon: UPDATE: I found that the 1460 is a well-defined value: https://cloud.google.com/kubernetes-engine/docs/concepts/network-overview

I updated the mtu value to 1460 (matching the parent mtu).
I also added a comment to explain that this was the value at the time of writing.

SgtCoDFish (approved):
/lgtm
/approve

maelvls:
I am also keen to know what led you to question the MTU. 😇

PR #928, opened 2023-11-15
Use latest make-dind images

inteon:
Use new debian 12 make-dind image.

see https://prow.build-infra.jetstack.net/view/gs/jetstack-logs/logs/ci-cert-manager-master-e2e-v1-23/1724781156580200448 for a successful e2e test that was run using this image (this image also fixes the buildx problem).


inteon:
/approve
/lgtm
self-lgtm'ed based on proven success, will revert if I missed something

PR #930, opened 2023-11-15
Re-add docker buildx

inteon:
It seems like we still have to additionally install buildx. Otherwise we get the following error:
```
DEPRECATED: The legacy builder is deprecated and will be removed in a future release.
            Install the buildx component to build images with BuildKit:
            https://docs.docker.com/go/buildx/
```
I tried to remove this line in https://github.com/cert-manager/testing/pull/927


inteon:
/approve
self-lgtm'ng to test for any additional issues with this image

PR #934, opened 2023-12-13
Update approver-policy test targets, optimized for the ci/cd refactor

inteon:
Linked to https://github.com/cert-manager/approver-policy/pull/306

### DIFF
--- a/config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
+++ b/config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
@@ -6,12 +6,15 @@ presubmits:
     always_run: true
     max_concurrency: 8
     annotations:
-      testgrid-create-test-group: 'false'
+      preset-go-cache: "true"
+      preset-local-cache: "true"
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/golang-dind:20231115-6d9432b-1.21.3
+      - image: eu.gcr.io/jetstack-build-infra-images/make-dind:20231115-8c0910c-bookworm
         args:
+        - runner
         - make
+        - vendor-go
         - verify
         resources:
           requests:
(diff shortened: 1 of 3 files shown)


config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
```diff
@@ -6,12 +6,15 @@ presubmits:
     always_run: true
     max_concurrency: 8
     annotations:
-      testgrid-create-test-group: 'false'
+      preset-go-cache: "true"
+      preset-local-cache: "true"
```
inteon, line 10: Enable go cache and download cache.

config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
```diff
@@ -6,12 +6,15 @@ presubmits:
...
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/golang-dind:20231115-6d9432b-1.21.3
+      - image: eu.gcr.io/jetstack-build-infra-images/make-dind:20231115-8c0910c-bookworm
         args:
+        - runner
         - make
+        - vendor-go
```
inteon, line 17: vendor go instead of using golang-dind

config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
```diff
@@ -23,13 +26,16 @@ presubmits:
...
     spec:
       containers:
-      - image: eu.gcr.io/jetstack-build-infra-images/golang-dind:20231115-6d9432b-1.21.3
+      - image: eu.gcr.io/jetstack-build-infra-images/make-dind:20231115-8c0910c-bookworm
         args:
+        - runner
         - make
-        - test
+        - vendor-go
+        - test-unit
```
inteon, line 38: use new test target name

config/jobs/cert-manager/approver-policy/cert-manager-approver-policy-presubmits.yaml
wallrj: This won't work if `make` is run with parallelism enabled:
 * https://www.gnu.org/software/make/manual/make.html#Parallel

It'll start on the `verify` target before finishing the `vendor-go` target.

And I assume that you will want to use parallelism to speed up the builds in future.

wallrj (changes requested):

inteon:
`vendor-go` should work with parallelism, because the `$(NEEDS_GO)` dependency also changes when you set the target.
This will make all places where go is used wait for go to be downloaded.

wallrj:
> `vendor-go` should work with parallelism, because the `$(NEEDS_GO)` dependency also changes when you set the target. This will make all places where go is used wait for go to be downloaded. https://github.com/cert-manager/makefile-modules/blob/d7a585bc28d562618c015079b8a6d088a0acd239/modules/tools/00_mod.mk#L182-L199

Ah, I didn't know that and I didn't believe it until I tested it locally and I can see that it does indeed seem to work.

/lgtm 
/approve

PR #939, opened 2024-01-04
Fix bug that causes caching to fail

inteon:
Typo should have been `$(pwd)` instead of `$(PWD)`.

example: https://prow.build-infra.jetstack.net/log?container=test&id=1742895913195540480&job=pull-cert-manager-master-e2e-v1-28

```
/usr/local/bin/runner: line 37: PWD: command not found
```


inteon:
self-approving and self-merging.

PR #943, opened 2024-01-06
istio-csr: Use new test target names (part of migration to makefile-modules)

inteon:
Once we merge this PR, the tests on https://github.com/cert-manager/istio-csr/pull/221 should start succeeding and it should be ready to be merged.

### DIFF
--- a/config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml
+++ b/config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml
@@ -4,17 +4,36 @@ presubmits:
   - name: pull-cert-manager-istio-csr-verify
     decorate: true
     always_run: true
-    max_concurrency: 8
-    annotations:
-      testgrid-create-test-group: 'false'
-    branches:
-    - ^main$
+    labels:
+      preset-go-cache: "true"
+      preset-local-cache: "true"
     spec:
       containers:
-      - image: golang:1.21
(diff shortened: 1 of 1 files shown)


config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml
```diff
@@ -4,37 +4,57 @@ presubmits:
...
   - name: pull-cert-manager-istio-csr-ca-rotation
     decorate: true
     always_run: true
-    branches:
-    - ^main$
     labels:
+      preset-go-cache: "true"
+      preset-local-cache: "true"
       preset-service-account: "true"
       preset-dind-enabled: "true"
```
SgtCoDFish, line 49: comment: It feels like if these presets are going to be common for all things using makefile-modules, then it might be worth having something like `preset-makefile-modules` in the future. Or maybe not. Certainly not a blocker for this though!

SgtCoDFish (approved):
/lgtm
/approve
/hold

The change in base image here makes me slightly nervous that this might break existing tests when this merges (although I've not tested that). I don't think that's a huge problem because I suspect we can move quickly once this merges and complete the shift to makefile-modules.

I've added a hold for that reason - so we can coordinate when all the PRs merge.

inteon:
I'll unhold the PR, since the https://github.com/cert-manager/istio-csr/pull/221 PR is only blocked on this (all tests are passing locally).
Once that PR gets merged, all tests will be working again.
/unhold

PR #944, opened 2024-01-06
Cleanup approver-policy prowjob yaml

inteon:
Removes `max_concurrency: 8` which was just copied blindly from existing prowjobs afaik.
Also reorders labels to match order of other configs in file.


SgtCoDFish (approved):
/lgtm
/approve

PR #945, opened 2024-01-08
Update cache logic so local and shared caches are merged

inteon:
I realised that the latest shared cache and the local cache can just be merged.
The new latest shared cache should be the latest shared cache + everything that was downloaded in this run.
This should increase the hit-rate of our cache by a lot.


inteon:
self-approving and lgtm'ng so we can quickly build a new image with these improvements.

PR #949, opened 2024-01-08
Cache: Speedup file transfer by disabling compression and using --inplace

inteon:
- disabling compression, because apparently it does not make sense when working with local disks (https://superuser.com/a/109788)
- using `--inplace` which speeds up transfers (https://superuser.com/a/109788) but is not necessary because we already have a "latest" cache copy mechanism

Future possible improvement: remove the "latest" cache copy mechanism and use remove the `--inplace` option.


inteon:
self lgtm'ing and approving

PR #959, opened 2024-01-22
Move issuer-lib to prow (part of makefile modules refactor)

inteon:
This PR will unblock https://github.com/cert-manager/issuer-lib/pull/75.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -97,12 +97,6 @@ branch-protection:
             contexts:
             - pull-trust-manager-verify
             - pull-trust-manager-smoke
-        issuer-lib:
-          required_status_checks:
-            contexts:
-            - lint
-            - test-e2e
-            - test-unit
         csi-lib:
           required_status_checks:
             contexts:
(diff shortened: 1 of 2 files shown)


config/config.yaml
```diff
@@ -97,12 +97,6 @@ branch-protection:
             contexts:
             - pull-trust-manager-verify
             - pull-trust-manager-smoke
-        issuer-lib:
-          required_status_checks:
-            contexts:
-            - lint
-            - test-e2e
-            - test-unit
```
SgtCoDFish, line 105: question: I'm surprised we don't require _any_ checks here. I'd expect to still require the tests but for the names to change. Why are these removed?

inteon: You only have to add these checks if they are non-prow checks.
prowjobs will be added automatically.

SgtCoDFish: That goes against what we're doing for other components (e.g. trust-manager above this)

Do you have some docs on that?

inteon: https://docs.prow.k8s.io/docs/jobs/#posting-github-status-contexts
and
https://docs.prow.k8s.io/docs/jobs/#requiring-job-statuses

SgtCoDFish: I'm with you - I'd not noticed that before. The checks are required because they have `always_run: true` and `optional` defaults to `false`.

So we _could_ clean up a bunch of the other required_status_checks in this file! Not required for this PR of course.

SgtCoDFish (approved):
/lgtm
/approve

Thanks!

PR #961, opened 2024-01-31
Replace the trivy tests for the ctl image

inteon:
Replace them with tests for startupapicheck in the newer releases.

### DIFF
--- a/config/prowgen/prowspecs/specs.go
+++ b/config/prowgen/prowspecs/specs.go
@@ -52,6 +52,9 @@ var knownBranches map[string]BranchSpec = map[string]BranchSpec{
 
 		e2eCPURequest:    "7000m",
 		e2eMemoryRequest: "6Gi",
+
+		// This older cert-manager release uses the ctl image to run the statupapicheck test
+		containerNames: []string{"controller", "acmesolver", "ctl", "cainjector", "webhook"},
 	},
 	"release-1.13": {
 		prowContext: &pkg.ProwContext{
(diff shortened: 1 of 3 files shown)


config/prowgen/prowspecs/specs.go
```diff
@@ -177,7 +190,13 @@ func (m *BranchSpec) GenerateJobFile() *pkg.JobFile {
 		m.prowContext.Periodics(pkg.E2ETestFeatureGatesDisabled(m.prowContext, kubernetesVersion, m.e2eCPURequest, m.e2eMemoryRequest), 24)
 	}
 
-	for _, container := range []string{"controller", "acmesolver", "ctl", "cainjector", "webhook"} {
+	// Apply the default set of container names if none have been specified
+	// TODO: this is the set that we want to migrate to in the future
+	if m.containerNames == nil {
+		m.containerNames = []string{"controller", "acmesolver", "startupapicheck", "cainjector", "webhook"}
+	}
```
SgtCoDFish, line 197: nitpick: Do you think it'd be easier to edit this in the future if the `master` branch in `config/prowgen/prowspecs/specs.go` also specified its container names in that file, rather than defaulting here? I'm fine with the defaulting, but it's a little jarring to see `master` be a special case.

SgtCoDFish (approved):
/lgtm
/approve
/hold

Hold for the nitpick but it doesn't matter, happy to merge!

config/prowgen/prowspecs/specs.go
inteon: The goal is that v1.15, 1.16 etc. will also not have this option defined.
So it gradually becomes unused and can be removed in the (far) future.

wallrj (approved):
Thanks Tim.

/lgtm

Fixes https://github.com/cert-manager/cert-manager/issues/6674

PR #963, opened 2024-02-17
Disallow force pushes and deletions for protected branches

inteon:
To improve the security of our GitHub repos, we want to make sure that force pushes and deletions are disallowed for the protected branches.


maelvls:
For anyone else looking at this, the purpose of this PR is to reflect what's already configured in GitHub settings. For example, cert-manager/cert-manager was already configured with these settings:

Screenshot 2024-02-19 at 11 01 44

/lgtm
/approve

PR #965, opened 2024-02-20
Replace remaining jetstack names with cert-manager

inteon:
Replaces some of the remaining jetstack names with "cert-manager".


SgtCoDFish (approved):
/lgtm
/approve

Untested but happy to review a followup if needed!

PR #967, opened 2024-02-20
Enable 'Require a pull request before merging'

inteon:
New settings:
image

Works thanks to this upstream change: https://github.com/kubernetes/test-infra/pull/32038

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -53,6 +53,15 @@ branch-protection:
       # Disallow deletion of the protected branch.
       allow_deletions: false
       
+      # In the GH UI, the following 'required_pull_request_reviews' configuration
+      # results in checking the 'Require a pull request before merging' checkbox,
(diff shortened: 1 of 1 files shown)


config/config.yaml
```diff
@@ -52,6 +52,9 @@ branch-protection:
       allow_force_pushes: false
       # Disallow deletion of the protected branch.
       allow_deletions: false
+      
+      required_pull_request_reviews:
+        required_approving_review_count: 0
```
wallrj, line 57: Add a comment, explaining why we set this to 0.
Also explain why it is necessary to use this setting in addition to "enforce_admins" which is already documented as:
> Prevent administrators from accidentally pushing directly to protected branches

wallrj (changes requested):

wallrj (approved):
Thanks @inteon 

/lgtm

PR #982, opened 2024-04-16
Use correct url for new infra image repository

inteon:
Google artifact registry uses a different image url format.


inteon:
self-lgtm'ng since it is a small bugfix

PR #985, opened 2024-04-16
Don't fix the go patch version

inteon:
This will make the autobump job automatically upgrade all go images to the latest built patch version.
eg. 20240416-5ccb4a8-1.22.1 won't get auto-upgraded to 20240416-63d3df9-1.22.2
however, 20240416-5ccb4a8-1.22 does get auto-upgraded to 20240416-63d3df9-1.22


inteon:
example of such a auto-upgrade PR: https://github.com/cert-manager/testing/pull/986

PR #988, opened 2024-04-17
Fix failing cherrypicker deployment

inteon:
While that cherrypicker accepts a GH app token as authentication, it doesn't work properly.
Logically, it makes sense that the GH app does not work: the cherrypicker has to create a fork to create the cherrypick PR and thus needs to be able to create a fork (which is something a GH app cannot do).

### DIFF
--- a/prow/cluster/hook_deployment.yaml
+++ b/prow/cluster/hook_deployment.yaml
@@ -48,6 +48,7 @@ spec:
         - --github-app-private-key-path=/etc/github/cert
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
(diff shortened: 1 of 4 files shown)


prow/cluster/hook_deployment.yaml
```diff
@@ -48,6 +48,7 @@ spec:
         - --github-app-private-key-path=/etc/github/cert
         - --config-path=/etc/config/config.yaml
         - --job-config-path=/etc/job-config
+        - --hmac-secret-file=/etc/webhook/hmac
```
inteon, line 51: Unrelated change: make the use of the `/etc/webhook/hmac` file explicit.

PR #991, opened 2024-04-17
Use the new gencred image to automatically create the worker clusters' kubeconfigs

inteon:
Adds a gencred deployment that dynamically generates the worker clusters' kubeconfig secrets.


prow/Makefile
inteon, line 1: This file has unrelated changes that cleanup all the make targets that we use to bootstrap and update the prow kubernetes cluster.

PR #1003, opened 2024-04-18
Change periodic cert-manager jobs to use cron

SgtCoDFish:
This spreads the jobs pretty well throughout the day to try and minimise the amount of compute we're using for any one instant of cert-manager tests.

This algorithm seems to give reasonable performance in spread without spending too much time digging into making it perfect.

I've attempted to comment it as much as possible because I know it's likely to be harder to read than it was to write.

### DIFF
--- a/config/prowgen/pkg/context.go
+++ b/config/prowgen/pkg/context.go
@@ -19,8 +19,7 @@ package pkg
 
 import (
 	"fmt"
-	"strconv"
-	"time"
+	"math"
 )
 
 // ProwContext holds jobs and information required to configure jobs for a given release channel.
@@ -51,8 +50,6 @@ type ProwContext struct {
 
 	presubmits []*PresubmitJob
 	periodics  []*PeriodicJob
-
-	minutesCounter time.Time
 }
 
 // RequiredPresubmit adds a presubmit which is run by default and required to pass before a PR can be merged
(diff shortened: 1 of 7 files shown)


config/prowgen/pkg/context.go
```diff
@@ -156,19 +210,6 @@ func (pc *ProwContext) periodicDashboardName() string {
...
-// increases the next value returned. This helps to prevent every test running at the same
-// minute within the hour causing a spiky distribution of tests.
-func (pc *ProwContext) minutesValue() int {
-	minuteVal := pc.minutesCounter.Minute()
-
-	pc.minutesCounter = pc.minutesCounter.Add(4 * time.Minute)
-
-	return minuteVal
+func cronSchedule(minute int, startHour int, periodicityHours int) string {
+	return fmt.Sprintf("%02d %02d-23/%02d * * *", minute, startHour, periodicityHours)
```
inteon, line 214: For my information: how can the following be expressed?
A job that should run every 5 hours starting from hour 1.
eg.:
01:00:00
06:00:00
11:00:00
16:00:00
21:00:00

02:00:00
07:00:00
12:00:00
17:00:00
22:00:00

03:00:00
08:00:00
13:00:00
18:00:00
23:00:00

...

inteon: This is interesting: https://stackoverflow.com/a/70618521
Seems like the only simple solution is to have the cron run at 0, 5, 10, 15, 20 with 4 hours between the last hour and the next day.

So the following expressions would resolve to the following sequences:
0-23/5 -> 0, 5, 10, 15, 20 (equivalent to */5)
1-23/5 -> 1, 6, 11, 16, 21
2-23/5 -> 2, 7, 12, 17, 22
3-23/5 -> 3, 8, 13, 18, 23
Next sequence == 0-23/5.

config/prowgen/pkg/context.go
```diff
@@ -116,18 +113,75 @@ func (pc *ProwContext) Periodics(job *Job, periodicityHours int) {
...
 			},
 		},
-		Interval: strconv.Itoa(periodicityHours) + "h",
-		// TODO: use Cron instead of Interval
-		// Cron: pc.cronSchedule(periodicityHours),
+		PeriodicityHours: periodicityHours,
+		// Use a minute and startHour of 0 when adding the period here, but in JobFile
+		// when we actually generate the JobFile struct we'll recalculate values
+		// to spread the periodics evenly across every hour / day
+		Cron: cronSchedule(0, 0, periodicityHours),
```
inteon, line 120: Can we leave the `Cron` value empty here? So it is clear that this value is not used.

config/prowgen/pkg/context.go
```diff
@@ -116,18 +113,75 @@ func (pc *ProwContext) Periodics(job *Job, periodicityHours int) {
...
+	for _, p := range pc.periodics {
+		periodicityCounts[p.PeriodicityHours] += 1
+	}
+
+	// periodicitySeen is used to track how many times we've seen a job with a given
+	// periodicity, which allows us to spread jobs with the same periodicity across the day
+	periodicitySeen := map[int]int{}
+
+	for i, p := range pc.periodics {
+		minute := (i * minuteSpread) % 60
```
inteon, line 148: Currently, all branches will start at the same minute and hour, could we maybe add some kind of hash to "seed" the minute and hour calculation?

inteon:
Thanks @SgtCoDFish. Took me some time to understand everything (had to refresh some of my cron-knowledge).
I have 2 remarks/ questions, but the PR seems to be good to go as-is, I think.
/approve
/lgtm
/hold in case you want to still address those two comments.

config/prowgen/pkg/context.go
SgtCoDFish: Meant to add some context here as to what I did so I'll do that now:

I went with a counter of the number of files we'd generated. Using a hash to seed RNG or something like that runs the risk of creating collisions with ~4 branches we're processing and the small numbers we're using, so there's a decent chance we'd end up no better off than having every branch running at the same minute. With a counter, it should differ a bit more!

PR #1008, opened 2024-04-19
With the new Makefile modules targets, we can more logically split the test Prowjobs

inteon:
I propose we have 3 jobs:
- verify
- unit+integration tests
- e2e tests

Previously, we had a seperate helm verify job and the unit, integration and verify targets were all performed together.
This causes some very confusing test results (the prow jobs shows that all unit tests succeeded, but the prowjob failed because the verify target failed)

### DIFF
--- a/config/prowgen/pkg/generators.go
+++ b/config/prowgen/pkg/generators.go
@@ -22,9 +22,78 @@ import (
 	"strings"
 )
 
-// MakeTest generates a test which runs linting and verification targets as well as
-// unit and integration tests
+// MakeTest generates a test which runs unit and integration tests
 func MakeTest(ctx *ProwContext) *Job {
+	job := jobTemplate(
+		"make-test",
+		"Runs unit and integration tests",
+		addLocalCacheLabel,
+		addGoCacheLabel,
+		addMaxConcurrency(8),
+	)
+
+	makeJobs, cpuRequest := calculateMakeConcurrency("2000m")
+
+	job.Spec.Containers = []Container{
+		{
+			Image: ctx.Image,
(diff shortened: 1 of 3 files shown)


config/prowgen/pkg/generators.go
```diff
@@ -35,6 +35,77 @@ func MakeTest(ctx *ProwContext) *Job {
...
+					Memory: "4Gi",
+				},
+			},
+		},
+	}
+
+	return job
+}
+
+// MakeTestOld generates a test which runs linting and verification targets as well as
```
ThatsMrTalbot, line 60: Nit: incorrect comment here, copy paste error

inteon:
> `make ci-presubmit` used to set `-j1` explicitly, we now set `-j2` on verify. Was there a reason concurrency was limited originally?

I think that by principle, the makefiles should be 100% parallelizable.
The limitation was just there to optimize the CPU - Memory ratio of the test to optimize our server utilisation (see https://github.com/cert-manager/testing/blob/master/config/prowgen/pkg/util.go#L31).

Having said that, I do think there are some parallelisation bugs in the current makefiles.
I think we should first try to fix those bugs in cm/cm and only restrict parallelisation if we determine that it is not possible to fix the issues there.

inteon:
Well, that caused a LOT of github notifications.
Seems like my comment here https://github.com/cert-manager/testing/pull/1008#issuecomment-2072221520 was seeing things a bit too much through rose-colored glasses...
I applied a quick-fix (instead of calling the verify command, we now call ci-presubmit which limits the concurrency of the Make target.

PR #1014, opened 2024-04-22
Restrict branchprotection branches, which are regexes

inteon:
The branch protection include `main` also included branches named `self-upgrade-main`.
This is unintentional and can be fixed by adding ^ and $ to the start and end of the regex.


SgtCoDFish (approved):
/lgtm
/approve

PR #1022, opened 2024-05-03
Switch from nginx to gke ingress and add annotation to use static ip

inteon:
See https://github.com/cert-manager/infrastructure/pull/41 for the infrastructure change.

### DIFF
--- a/prow/cluster/deck_service.yaml
+++ b/prow/cluster/deck_service.yaml
@@ -15,10 +15,12 @@
 apiVersion: v1
 kind: Service
 metadata:
-  labels:
-    app: deck
   namespace: default
   name: deck
+  labels:
+    app: deck
+  annotations:
+    cloud.google.com/neg: '{"ingress": true}'
 spec:
   selector:
     app: deck
--- a/prow/cluster/tls-ing_ingress.yaml
+++ b/prow/cluster/tls-ing_ingress.yaml
@@ -15,14 +15,15 @@
 apiVersion: networking.k8s.io/v1
 kind: Ingress
 metadata:
+  name: prow-ing
+  namespace: default
   annotations:
     cert-manager.io/cluster-issuer: letsencrypt-prod
     kubernetes.io/tls-acme: "true"
     acme.cert-manager.io/http01-edit-in-place: "true"
-  name: prow-ing
-  namespace: default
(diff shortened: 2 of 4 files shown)


prow/cluster/deck_service.yaml
```diff
@@ -15,10 +15,12 @@
...
 kind: Service
 metadata:
-  labels:
-    app: deck
   namespace: default
   name: deck
+  labels:
+    app: deck
+  annotations:
+    cloud.google.com/neg: '{"ingress": true}'
```
inteon, line 23: This annotation is required because this Service is a ClusterIP.

prow/cluster/tls-ing_ingress.yaml
```diff
@@ -15,14 +15,15 @@
...
 metadata:
+  name: prow-ing
+  namespace: default
   annotations:
     cert-manager.io/cluster-issuer: letsencrypt-prod
     kubernetes.io/tls-acme: "true"
     acme.cert-manager.io/http01-edit-in-place: "true"
-  name: prow-ing
-  namespace: default
+    kubernetes.io/ingress.class: "gce"
```
maelvls, line 24: **Unrelated:** GCP still doesn't support `ingressClassName`... and OpenShift 4.12 onwards will spam you of alerts saying that you can have an ingress without `ingressClassName` set... the Kubernetes ecosystem is weird

maelvls (approved):

prow/cluster/tls-ing_ingress.yaml
inteon: GCP is really lacking here. It almost seems like they want you to go to Gateway API instead of fixing this.

prow/cluster/deck_service.yaml
maelvls: For anyone wondering why this annotation is set:

> In the Service manifest, you must use `type: NodePort` unless you're using [container native load balancing](https://cloud.google.com/kubernetes-engine/docs/concepts/container-native-load-balancing). If using container native load balancing, use the `type: ClusterIP`.

For context,

```yaml
cloud.google.com/neg: '{"ingress": true}'
```

is the annotation used to tell ingress-gce to enable container native load balancing.

PR #1032, opened 2024-06-13
Bump version of prow-controller-manager-spot

inteon:
Already applied to cluster (using `cd prow/ && make deploy-prow`)


inteon:
self-lgtm'ing since this is just a version bump

PR #1035, opened 2024-07-03
Add testing configuration for openshift-routes project

inteon:
See https://github.com/cert-manager/openshift-routes/pull/60 for implementation of test Make targets.


inteon:
self-lgtm'ed since the https://github.com/cert-manager/openshift-routes/pull/60 PR got merged

PR #1036, opened 2024-07-03
Update tide config to allow auto-merging github-actions skip-review PRs

inteon:
If the PR's author is `github-actions[bot]` and the PR has the `skip-review` label, we can merge the PR automatically (without review) after the tests pass.
Also fixes an issue for cert-manager/cert-manager which had overlapping tide queries.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -210,6 +210,8 @@ tide:
   # Default tide config for all repos in the cert-manager org
   - orgs:
     - cert-manager
+    excludedRepos:
+    - cert-manager/cert-manager # Handled with a separate Tide query below.
     labels:
     - lgtm
     - approved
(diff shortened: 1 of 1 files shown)


config/config.yaml
```diff
@@ -210,6 +210,8 @@ tide:
   # Default tide config for all repos in the cert-manager org
   - orgs:
     - cert-manager
+    excludedRepos:
+    - cert-manager/cert-manager # Handled with a separate Tide query below.
```
inteon, line 214: This should fix tide for cert-manager/cert-manager, where it was running two queries instead of one.

SgtCoDFish (approved):
/lgtm
/approve

I think this is worth a go! Thanks for doing it!

PR #1044, opened 2024-07-16
Remove unused K8S_VERSION environment variable from istio-csr tests

inteon:
Currently, the K8S_VERSION environment variable is ignored when running istio-csr tests.
To avoid confusion, I propose to not set the variable in the tests and rename the tests so it is clear they all just use the kubernetes version defined in the repo.

We might want to add support for configurable K8S_VERSION in the future. The K8S_VERSION defined today are however just random and unuseful.


SgtCoDFish (approved):
/lgtm
/approve

Thanks 👍

PR #1045, opened 2024-07-16
Add skip-review label to label-sync config

inteon:
Based on https://github.com/kubernetes/test-infra/blob/f4dfb8f3f0062895f1d66a659f52c91409c45ef9/label_sync/labels.yaml#L1109-L1113

This will add the label to the known labels on all repositories.

### DIFF
--- a/config/labels.yaml
+++ b/config/labels.yaml
@@ -404,3 +404,8 @@ default:
     name: spam
     target: prs
     addedBy: member or author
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

Untested but happy to review a followup if this doesn't work. It obviously looks correct though!

PR #1049, opened 2024-07-17
Auto-run tests on PRs created by github-actions

inteon:
Add `github-actions` to the list of `trusted_apps`, such that tests are run automatically.

Also simplifies the `repos` section of the triggers configuration.

Worked for https://github.com/cert-manager/approver-policy/pull/468.

### DIFF
--- a/config/plugins.yaml
+++ b/config/plugins.yaml
@@ -4,13 +4,9 @@
 ---
 triggers:
 - repos:
-  - cert-manager/testing
-  - cert-manager/cert-manager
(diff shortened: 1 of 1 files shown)


SgtCoDFish (approved):
/lgtm
/approve

I'm a little wary of the change to use the org rather than naming specific repos - my initial reaction is "what about stuff like cert-manager/infrastructure". But I suspect it just won't be relevant and it's easy to change back if needed!

Thanks for doing this!

PR #1050, opened 2024-07-19
Add test-e2e-ambient for 1.22 release

paulwilljones:
⚠️ Only merge after https://github.com/cert-manager/istio-csr/pull/336

### DIFF
--- a/config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml
+++ b/config/jobs/cert-manager/istio-csr/cert-manager-istio-csr-presubmits.yaml
@@ -288,7 +288,65 @@ presubmits:
             memory: 6Gi
         env:
         - name: ISTIO_VERSION
-          value: "1.22.2"
+          value: "1.22.3"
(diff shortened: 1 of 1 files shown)


inteon:
@SgtCoDFish do we still have to merge this? Also, is the `test-e2e-pure-runtime` target being called anywhere already?
A cleaner solution might be to bundle these targets in the istio-csr repo under `test-e2e` and just keep calling a single Makefile target here.

SgtCoDFish:
Yeah this needs to merge I think. I don't think it's the end of the world that it hasn't gone in yet though!

I'm open to reworking how these tests are called down the road as long as each e2e test can still easily be run individually!

SgtCoDFish:
Oh, maybe this should run as a separate test rather than being tacked onto an existing one 🤔  I'll sort that out

SgtCoDFish:
@inteon I've changed this to add pure runtime tests + ambient tests as separate targets 👍

inteon (approved):
/lgtm
/approve

PR #1055, opened 2024-09-05
Add note explaining where a required check is defined

SgtCoDFish:
In https://github.com/cert-manager/testing/pull/789 we removed the presubmit from this repo, but the test is
still marked as required for the website repo.

This test is defined in the website repo instead.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -110,6 +110,10 @@ branch-protection:
           required_status_checks:
             contexts:
             - netlify/cert-manager/deploy-preview # See https://github.com/cert-manager/infrastructure#netlify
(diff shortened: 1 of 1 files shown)


wallrj (approved):
Thanks @SgtCoDFish 

Sorry for not giving the GitHub action a more distinctive name.
I kind of knew it would cause confusion, but it seemed like the easy thing to do at the time.

/approve
/lgtm

SgtCoDFish:
/unhold

I'm fine with the name, just think this needs documenting clearly because we run the website tests differently to how we run presubmits elsewhere!

PR #1057, opened 2024-09-17
Fix incorrect http01 acme solver configuration

inteon:
The HTTP01 solver was still configured to solve challenges using ingress instead of Gateway API.


SgtCoDFish (approved):
/lgtm
/approve

DId this _ever_ work? 😁  Thank you for looking into it!

PR #1060, opened 2024-09-26
Add k8s 1.31 test for cert-manager 1.15

SgtCoDFish:
Follows on from https://github.com/cert-manager/cert-manager/pull/7305

### DIFF
--- a/config/prowgen/prowspecs/specs.go
+++ b/config/prowgen/prowspecs/specs.go
@@ -109,7 +109,7 @@ var knownBranches map[string]BranchSpec = map[string]BranchSpec{
 		primaryKubernetesVersion: "1.30",
 
 		// TODO: test k8s 1.31 here when possible; requires support in the release-1.15 branch on cert-manager
-		otherKubernetesVersions: []string{"1.25", "1.26", "1.27", "1.28", "1.29"},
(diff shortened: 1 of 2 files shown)


wallrj (approved):
A couple of  non-blocking questions and a suggestion.

/approve
/lgtm

/hold in case you want to address those here.

config/prowgen/prowspecs/specs.go
```diff
@@ -109,7 +109,7 @@ var knownBranches map[string]BranchSpec = map[string]BranchSpec{
 		primaryKubernetesVersion: "1.30",
 
 		// TODO: test k8s 1.31 here when possible; requires support in the release-1.15 branch on cert-manager
-		otherKubernetesVersions: []string{"1.25", "1.26", "1.27", "1.28", "1.29"},
+		otherKubernetesVersions: []string{"1.25", "1.26", "1.27", "1.28", "1.29", "1.31"},
```
wallrj, line 112: Ah, I'd forgotten that all the YAML is generated  these days.
Perhaps that's why the cron schedules changed.

wallrj:
/unhold

I'll fix that comment in https://github.com/cert-manager/testing/pull/1061

config/prowgen/prowspecs/specs.go
SgtCoDFish: It is, yeah!

PR #1065, opened 2024-10-16
Don't test master against k8s 1.27

SgtCoDFish:
By the time we release cert-manager 1.17 in ~Feb 2025, all cloud providers will have dropped support for Kubernetes 1.27 except OpenShift.

I think we should drop 1.27 as a tested version but keep it as a supported version, as per the recently added documentation on our supported releases page: https://cert-manager.io/docs/releases/#supported-vs-tested-versions-of-kubernetes


inteon:
I agree, we don't have to test these old versions.

/approve
/lgtm

PR #1066, opened 2024-10-22
Add Istio 1.24 test

howardjohn:
See https://github.com/cert-manager/istio-csr/pull/422 for context


SgtCoDFish:
/ok-to-test

Thank you for raising this! I've taken this as motivation to fix up weirdness around how we test istio versions: https://github.com/cert-manager/istio-csr/pull/429

Once that PR merges, we should be good to go on this one I think!

SgtCoDFish (approved):
/lgtm
/approve

Thank you for this! 🚀

PR #1067, opened 2024-10-25
istio-csr Test Cleanup / Addition

SgtCoDFish:
1. Removes old versions of istio (all are EOL and not supported since at least Apr 2024)
2. Adds a periodic test of istio-csr against the latest version of istio. I've not been able to e2e test the syntax I've used but I've roughly been able to approximate prow locally and it seemed OK.


inteon (commented):
/approve
/lgtm

PR #1068, opened 2024-10-28
Add dashboard group for istio-csr, fix comment

SgtCoDFish:
- Adds a couple of annotations which seem obvious (see https://github.com/kubernetes/test-infra/blob/master/testgrid/config.md)
- Fixes times given in comment on cron field; I mistakenly used BST rather than UTC
-  Moves the new istio-csr dashboard to a group called "cert-manager-subprojects"

### DIFF
--- a/config/testgrid/dashboards.yaml
+++ b/config/testgrid/dashboards.yaml
@@ -8,6 +8,9 @@ dashboard_groups:
   - cert-manager-periodics-release-1.16
   - cert-manager-presubmits-master
   - cert-manager-testing-janitors
+- name: cert-manager-subprojects
+  dashboard_names:
+  - istio-csr-periodics
 
 # Dashboards
 dashboards:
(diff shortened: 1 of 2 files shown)


config/testgrid/dashboards.yaml
```diff
@@ -9,6 +9,11 @@ dashboard_groups:
   - cert-manager-presubmits-master
   - cert-manager-testing-janitors
 
+dashboard_groups:
```
inteon, line 12: It seems like there is a duplicate key here, is this file auto-generated?

SgtCoDFish: Apologies, just an oversight on my end! Fixed 👍

inteon:
The PR looks good to me, I think the best way to figure out if it will work is to merge it.
/approve
/lgtm

SgtCoDFish:
It worked! https://testgrid.k8s.io/cert-manager-subprojects

PR #1079, opened 2025-03-27
Upgrade image variants

inteon:
Bumping a few of the image variant parameters.
Next, image builds will get triggered and the autobumper will bump these images in the configurations.

### DIFF
--- a/images/golang-dind/build.yaml
+++ b/images/golang-dind/build.yaml
@@ -1,14 +1,10 @@
 name: golang-dind # Name of the image to be built
 
 variants:
-  "1.22":
+  "1.24":
     arguments:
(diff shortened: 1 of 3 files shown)


images/golang-dind/build.yaml
```diff
@@ -1,14 +1,10 @@
 name: golang-dind # Name of the image to be built
 
 variants:
-  "1.22":
+  "1.24":
```
inteon, line 4: I don't think we have any location where we HAVE to use an older go version, just reduced the number of variants to just 1.24 here. The old images will remain in the registry btw.
The goal is to update all places where this image is used to 1.24 after merging this PR.

SgtCoDFish: Sounds good to me, we should adopt 1.24 everywhere soon anyway!

SgtCoDFish (approved):
/lgtm
/approve

PR #1092, opened 2025-05-02
fix invalid color in new label

SgtCoDFish:
Should fix https://storage.googleapis.com/cert-manager-prow-artifacts/logs/periodic-testing-label-sync/1918293832294731776/build-log.txt

### DIFF
--- a/config/labels.yaml
+++ b/config/labels.yaml
@@ -409,7 +409,7 @@ default:
     name: skip-review
     target: prs
     addedBy: autobump bot
-  - color: 006699
+  - color: "006699"
     description: Indicates that a PR should be squashed by tide
     name: tide/squash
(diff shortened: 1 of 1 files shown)


wallrj (approved):
Did you see that there's already another squash label ?

https://github.com/cert-manager/testing/blob/287498b2e97d529788a3a05334b0be2aef08e4a8/config/labels.yaml#L378-L384

SgtCoDFish:
Yeah but we set the label to this value explicitly: https://github.com/cert-manager/testing/blob/287498b2e97d529788a3a05334b0be2aef08e4a8/config/config.yaml#L218

(as this is approved and trivial, I'll manually add the lgtm label now)

PR #1098, opened 2025-07-04
Add config for cert-manager/webhook-cert-lib repo

inteon:
See https://github.com/cert-manager/webhook-cert-lib


erikgb (commented):
/lgtm
/approve
/hold
Just in case you want to wait.

PR #1101, opened 2025-07-06
Initial Setup for go-pkcs12

ali-hamza-noor:
Setting up the testing for go-pkcs12 fork

### DIFF
--- a/config/jobs/cert-manager/go-pkcs12/go-pkcs12-presubmits.yaml
+++ b/config/jobs/cert-manager/go-pkcs12/go-pkcs12-presubmits.yaml
@@ -1,0 +1,24 @@
+presubmits:
+  cert-manager/go-pkcs12:
+
+    - name: pull-cert-manager-go-pkcs12-verify
+      always_run: true
+      max_concurrency: 8
+      decorate: true
+      branches:
+        - master
+      spec:
+        containers:
+          - image: europe-west1-docker.pkg.dev/cert-manager-tests-trusted/cert-manager-infra-images/golang-dind:20250327-a3af8ba-1.24
+            args:
+              - make
+              - test
+            resources:
+              requests:
+                cpu: 2
+                memory: 4Gi
+        dnsPolicy: None
+        dnsConfig:
+          nameservers:
+            - 8.8.8.8
+            - 8.8.4.4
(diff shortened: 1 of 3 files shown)


config/jobs/cert-manager/go-pkcs12/go-pkcs12-presubmits.yaml
```diff
@@ -0,0 +1,25 @@
...
+      max_concurrency: 8
+      decorate: true
+      branches:
+        - master
+      spec:
+        containers:
+          - image: europe-west1-docker.pkg.dev/cert-manager-tests-trusted/cert-manager-infra-images/golang-dind:20250327-a3af8ba-1.24
+            args:
+              - runner
+              - make
```
inteon, line 15: @ali-hamza-noor, I don't think there is a Makefile in that repo.

erikgb: Merged now, so we have a Makefile.

config/jobs/cert-manager/go-pkcs12/go-pkcs12-presubmits.yaml
```diff
@@ -0,0 +1,25 @@
...
+              - test
+            resources:
+              requests:
+                cpu: 2
+                memory: 4Gi
```
erikgb, line 20: Maybe we can reduce this request for resources?

triage_party/triageparty_configmap.yaml
```diff
@@ -50,6 +50,7 @@ data:
         - https://github.com/cert-manager/klone
         - https://github.com/cert-manager/boilersuite
         - https://github.com/cert-manager/google-cas-issuer
+        - https://github.com/cert-manager/testing
```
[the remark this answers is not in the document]
erikgb:
```suggestion
        - https://github.com/cert-manager/go-pkcs12
```

Maybe?

config/jobs/cert-manager/go-pkcs12/go-pkcs12-presubmits.yaml
erikgb: Dunno, maybe just keep it - if we doesn't pay for the resource consumption. I cannot imagine a simple go test setup requires 2 cores and 4Gi of memory. 😆

wallrj (approved):
Thanks @ali-hamza-noor 

I think this looks fine. Similar to https://github.com/cert-manager/testing/pull/956 where Tim added the helm-tool repo to prow.

My only question is whether there's something else we need to do to configure prow to set the branch protection...it looks like that's already configured for the whole cert-manager GitHub org, but if not we can add that in another PR.

/approve
/lgtm

PR #1108, opened 2025-09-12
Preparing k8s v1.34 test suite

hjoshi123:
Adds optional tests so that we can test our CI for k8s 1.34 before making it default


erikgb:
@hjoshi123, I don't think you are supposed to manually edit `cert-manager-master.yaml`. Run `make prowgen` to fix CI.

hjoshi123:
Ah gotcha. Let me try that. Thank you @erikgb

PR #1113, opened 2025-10-02
Renovate tested Istio version in istio-csr

erikgb:
While working on istio-csr, I noticed that we:

- test a lot of Istio versions that are EOL
- not using the latest patch for all minors
- missing tests for newer versions

I suggest removing tests for Istio releases that are EOL, which is most of them.

https://istio.io/latest/docs/releases/supported-releases/#support-status-of-istio-releases


erikgb:
/hold

Adding a hold as I think we should remove some older versions, ideally all versions that are EOL now.

PR #1183, opened 2026-05-25
Replace gencred with workload identities

inteon:
Similar to upstream (https://github.com/kubernetes/test-infra), we can stop using gencred and just use workload identities directly intead.

depends on https://github.com/cert-manager/infrastructure/pull/77


SgtCoDFish (approved):
/lgtm
/approve

Thank you for this, I imagine this went much faster because you did it!

PR #1185, opened 2026-05-26
Connect to k8s using DNS endpoint

inteon:
See https://github.com/cert-manager/infrastructure/pull/78

Changes kubeconfigs to use DNS endpoint instead of static IP.


prow/cluster/kubeconfig-prow-trusted_configmap.yaml
```diff
@@ -13,8 +13,7 @@ data:
...
+        server: https://gke-f93ae377867f444e8403fe74521322a59498-264813140030.europe-west1-b.gke.goog
```
SgtCoDFish, line 16: note: the cert presented here is publicly trusted!

SgtCoDFish (approved):
/lgtm
/approve

Thank you!

prow/cluster/kubeconfig-prow-trusted_configmap.yaml
inteon: Indeed, this is one of the "features" of this google-owned DNS endpoint.

PR #1225, opened 2026-08-12
Strip live credentials from presubmit jobs

wallrj:
Presubmits run unreviewed pull-request code, so they should never mount live credentials. This hardens the Prow config so that only periodics and postsubmits (which run merged, reviewed code) carry credential presets.

### Changes

- **Strip credential presets at the choke point.** `addPresubmit()` in `config/prowgen/pkg/context.go` now deletes the `preset-venafi-*` presets from every presubmit it registers. That is the single point through which every presubmit passes, so a future generator that re-adds a credential label to a presubmit cannot reintroduce the problem. New unit tests in `context_test.go` assert presubmits are stripped while periodics retain their credentials.
- **Remove the per-issuer Venafi presubmits.** The `pull-*-issuers-venafi-{tpp,cloud,ngts}` jobs were optional, `/test`-only presubmits that never ran by default; with the presets stripped they could no longer authenticate, so triggering one would only fail. Their generators, label configurers and the now-orphaned per-issuer `GINKGO_FOCUS` presets are removed too.
- **Censor secrets from job logs.** `censor_secrets: true` in the default decoration config, so the sidecar redacts known secret values before logs and artifacts are uploaded to the public `cert-manager-prow-artifacts` bucket. Defence in depth alongside removing the presets.

### Coverage

Regular Venafi coverage is unaffected: it comes from the `ci-*-issuers-venafi` periodic, which keeps its credentials. Periodics and postsubmits retain all credential presets.

### Validation

- `go test ./config/prowgen/...`
- `make prowgen` produces no drift
- `make local-checkconfig` (`checkconfig --strict`) passes

### Review fixes

- **Mount the Venafi Secrets as volumes in their presets.** Prow's censoring only learns the values to redact from Secrets *mounted* into the Pod, never from env vars, so `censor_secrets` was not actually covering the env-injected Venafi credentials. Each preset now also mounts its Secret (only the sensitive keys) under `/etc/censor/`; evidence permalinks pinned to the utility-image version in use are in the preset comment.
- **Repo-wide policy test.** `config/prowgen/presubmit_policy_test.go` forbids secret-bearing presets on any presubmit, hand-written or generated. The forbidden set is derived from the preset definitions (any preset injecting a Secret via env `secretKeyRef` or a Secret volume), so new credential presets are covered automatically. Runs in the existing `pull-testing-test` presubmit via `make test`.

### Follow-ups

- #1226 — pre-merge Venafi coverage is intentionally gone from this PR (a credentialed presubmit is exactly the vulnerability); the proposed substitute is a vendor-supplied verified fake of the Venafi APIs, which would let credential-free per-issuer presubmits return.
- cert-manager/infrastructure#89 adds a dedicated tainted node pool for credentialed jobs. Stripping presets stops presubmits *mounting* the Venafi credentials, but the `ci-*-issuers-venafi` periodics still share nodes with privileged (dind) presubmit pods, so a node escape could read their secrets via the kubelet. Once that pool exists, a follow-up here will pin the Venafi periodics to it with a `nodeSelector` and toleration.

*with claude opus-4.8 and claude fable-5*


SgtCoDFish (approved):
/lgtm
/approve

I can't see any reason to not merge!

PR #1228, opened 2026-08-14
Pin credentialed periodics to the dedicated node pool

wallrj:
Pins every periodic which carries a credential preset (currently the `*-issuers-venafi` periodics on master, release-1.20 and release-1.21) to the dedicated `dedicated=credentialed-jobs` node pool added by cert-manager/infrastructure#89, by injecting the matching `nodeSelector` and toleration in prowgen's `Periodics()` — the single point through which every generated periodic passes. This mirrors the shape of the presubmit credential guard in `addPresubmit()`: the next generated credentialed job cannot silently land back on the shared worker pool, where presubmits run unreviewed PR code in privileged (dind) pods and can poison the node-local hostPath build caches.

The taint key/value contract is defined in cert-manager/infrastructure `gcp/modules/gcp-cluster/main.tf`; the comment on `credentialPresets` in `pkg/context.go` points there.

Verified with `go test ./...`, `make verify` (regenerated output matches the committed files) and prow `checkconfig --strict` (passes).

**Hold until infrastructure#89 is merged *and applied*:** until the node pool exists, the injected `nodeSelector` matches no node and the Venafi periodics would sit Pending.

*with claude fable-5*

### DIFF
--- a/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
+++ b/config/jobs/cert-manager/cert-manager/master/cert-manager-master.yaml
@@ -620,6 +620,13 @@ periodics:
             nameservers:
                 - 8.8.8.8
                 - 8.8.4.4
+        nodeSelector:
+            dedicated: credentialed-jobs
+        tolerations:
+            - key: dedicated
+              operator: Equal
+              value: credentialed-jobs
+              effect: NoSchedule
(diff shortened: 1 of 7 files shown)


inteon (approved):

inteon (commented):
/approve
/lgtm

